PR #172, opened 2016-01-18
Proposing a few additions to "Apps Library" landing page

shawncrawley:
This was inspired by many people that said, after visiting apps.hydroshare.org, that they would like to know what apps did before clicking on them.  First, I added a 'description' attribute to the app.py_tmpl where a brief description of the app should be placed. I then added a question mark icon to all app icons on the Apps Library page that when hovered over will provide an overlay on the app icon showing the description of the app. In the case of a touch-device, the question mark must be clicked, upon which an X also appears with a click event to close it. Here is a screenshot of what it looks like:

image

image


sdc50:
This feature has actually been talked about for a while. Nathan had the
idea of the app description sliding over the app icon on hover. I have a
prototype of the app icons flipping over to show the description on the
back. I think this is definitely a feature we want. We just need to
agree/decide on the styling.

shawncrawley:
Awesome. I'm glad we're all on the same page then. Having the app icons flip over sounds cool. I'd vote for that.

swainn:
Shawn, I'm afraid you submitted the pull request on the master branch. Can you please get the latest version of the dev branch, add your changes, and submit a new pull request?

PR #173, opened 2016-01-18
Added app description help to Apps Library landing page

shawncrawley:
This was inspired by many people that said, after visiting apps.hydroshare.org, that they would like to know what apps did before clicking on them.  First, I added a 'description' attribute to the app.py_tmpl where a brief description of the app should be placed. I then added a question mark icon to all app icons on the Apps Library page that when hovered over will provide an overlay on the app icon showing the description of the app. In the case of a touch-device, the question mark must be clicked, upon which an X also appears with a click event to close it. Here is a screenshot of what it looks like:

image

image


shawncrawley:
@swainn I think all is good now. The pull request is from my forked dev branch.

sdc50:
I just pushed a branch called app-transitions. It adds this functionality but has the description on the back of the app icon with a nice transition to flip it over on hover. I think the code is functional but the icon takes a second to reappear when flipped back over, which I find slightly annoying and I haven't been able to fix. The branch is pushed so others can check it out. We now need to decide if:
* we merge this pull request
* we include the app-transition branch as is
* we dedicate more time to fine tune the styling on the app-transition branch
* we come up with some other way for displaying app descriptions

I think the first two options could be included in the 1.3 release. The latter two would take more time.

swainn:
Let me take a look at this this afternoon. One question I have is if your flipping is still causing issues with the footer like you mentioned before.

sdc50:
I can't remember exactly what the issues was, but I think that I fixed it.

swainn:
Ok. I finally got to take a look. I think the two options need to be combined (which is strange because one is derived from the other, but hear me out). 

@sdc50: the flip transition that you have added is great ux, but you also removed the "?" button, which was necessary for mobile users to view the description--we must support mobile. While having the app logo behind the text is cool, I'm concerned about readability and there needs to be some experimenting done there, so perhaps we leave that out on this iteration. Also, it seems unnecessary to have the app icon flip when there is no description.

@shawncrawley: I already mentioned I like the "?" button for giving access to both mobile and desktop users, however, it may be better to hide the "?" button when there is no description. The other problem with your solution is that the overflow on the overlay element is not set, so if the description is longer than would fit on the app tile, the text flows over outside of the tile. Not very classy.

So, what I think what we need is basically @sdc50's version that is triggered by hovering over a "?" button in the corner like @shawncrawley's version. Also, hide the app icon behind the text and make the "?" button only show up if there is a description. Thoughts?

sdc50:
Here are a few thoughts just for the sake of discussion.

The "?":
If voting I would give the ? a -0. I prefer the cleaner look of not having
them, but realize that it helps a user realize that there is additional
information there. It also provides a way for accessing it on touch screen
devices, which the card flip on hover doesn't currently have. My thought
would be to swipe over the icon to flip it on touch screen, but that is
getting beyond my comfort zone in CSS/JavaScript. The other challenge with
that is that it wouldn't be very intuitive that you need to swipe the icon
to see the description. A possible solution would be to have a CSS
transition of the icon flipping when the page loads, but that would of
course add complexity and development time.

Icon on the back:
I was playing with the opacity of the back panel on the icons to make the
text legible. I didn't test a lot of icons so maybe the level wasn't quite
right. It may be easier to do what Nathan is suggesting and just remove the
icon (although I do like the effect of having the icon there).

No description:
Handling the case when there is no description would require some more
thought and more code. Would could have a default description (like the app
name) just so there is consistency with the behavior of the icons.
Hopefully app developers will realize the benefit of having a description
and will start including it.

Timing:
It looks like none of the solutions that we have right now are ready to go.
Is this feature urgent enough to delay the release of 1.3? I don't think it
would take too much more effort to finish it up, but I don't have time that
I can dedicate to it.

swainn:
While I am a minimalist like you, we need something to indicate that there is more than you are seeing. Not so big on desktop, because it flips when I move my mouse over it, but for mobile, something needs to indicate that. So, a simple solution for now is the "?". I think what I would like even better though is the info symbol (i in a circle). I think that is customary. Bootstrap has one that could be added easily.

Icon:
I like icon on the back, but I would want to experiment with it more before it is standard. Since all it is is opacity, it would be easy enough to take out and then add later as an improvement.

No Description:
It seems like it would be fairly easy with the button triggered version that @shawncrawley has to just not show the button if the variable that is storing the description is empty. Right?

As I think about it, the flipping could probably be added later when we aren't delaying a release. I propose that we use @shawncrawley's version with the following tweaks:

1. Set the "overflow-y" property to "scroll" on the overlay element that contains the description to prevent unsightly text overflow when the description is long.
2. Replace the "?" symbol with bootstrap's info symbol (glyphicon glyphicon-info-sign). See this page for how to do it: http://getbootstrap.com/components/#glyphicons Bootstrap is built into Tethys, so you won't need to add the libraries, just the HTML markup.
3. (optional) If it is easy to do, make the info button conditional based on whether the description variable is empty or not. That way it will only appear on tiles with a description.

What do you think? @shawncrawley could you work these in?

sdc50:
Sounds like a good plan to me if Shawn is able to work on it.

shawncrawley:
@swainn @sdc50 That sounds great. I'll get to work on those 3 changes (#3 should actually be easy) and hopefully be done soon (today).

shawncrawley:
@swainn @sdc50 In fact, I think it is good to go. Give it a look.

swainn:
Excellent. One question: what is the behavior for mobile? With the mouse, the description hides when my mouse hovers off the app icon. How is the description hidden on mobile devices, since there is not hover-off to detect. 

One last thing before we can merge this: I noticed you made all the changes to the CSS files directly. Unfortunately, the css files are generated from the less files (located one level up in a directory called less). I'll need you to make those changes in the less files, regenerate the css files, and then commit both. Otherwise, your changes will get blown away the next time someone builds the less files. http://lesscss.org/

shawncrawley:
@swainn I think it is finally good. To answer your question, there is JavaScript that detects if the device has a touch screen: 

`if (('ontouchstart' in window) || (navigator.maxTouchPoints > 0) || (navigator.msMaxTouchPoints > 0))`

If so, the event is triggered on click, upon which an 'X' also appears that will close the app info.
If not, the event is triggered on mouseover.

sdc50:
Shawn,

Were you able to figure out the less files and how to recompile them? If
you have any questions about that I can drop by your office.

shawncrawley:
@sdc50 Do I need to recompile them before I commit and push? Could you come by now and show me?

sdc50:
Yes they should be recompiled just to make sure there is nothing in the CSS
that will get overwritten and surprise us later. I'm in the library now,
but will be over there in 15 min or so.

shawncrawley:
@sdc50 I believe I was able to perform the recompilation. You can still stop by just in case, or if it doesn't look right.

shawncrawley:
@swainn I did a few hasty commits/pushes before fully testing the code, so a few subsequent pushes were required. But now I am confident in the most recent commit. Sorry about that.

swainn:
I just performed the merge. Thank you, @shawncrawley, for your contribution. I think this will be a nice feature for Tethys apps.

PR #174, opened 2016-01-20
handle NotAllowedToDisconnect exception

zhiyuli:
@swainn @sdc50 
We are having an issue on apps.hydroshare.org: HydroShare users login apps.hydroshare.org tethys server through OAuth, tethys will create a native tethys user and associate it with the HydroShare social account. If users try to disconnect the native tethys user from HydroShare social account, it will raise a NotAllowedToDisconnect exception, which results in a 500 error. 

I am guessing this is because the native tethys user created by HS OAuth is somehow different from the regular tethys user that was originally created via sign-up page and then got associated with HydroShare social account. At least the former user has no native tethys password, so it is useless without a HydroShare social account.

This patch handles the NotAllowedToDisconnect exception and displays a message to the users. It is still a short-term solution. A long-term solution might be let the user create a native tethys password before disconnecting from any social account.

### DIFF
--- a/tethys_portal/middleware.py
+++ b/tethys_portal/middleware.py
@@ -12,7 +12,7 @@ from django.shortcuts import redirect
 from social.apps.django_app.middleware import SocialAuthExceptionMiddleware
 from django.http import HttpResponse
 from social import exceptions as social_exceptions
-from social.exceptions import AuthCanceled, AuthAlreadyAssociated
+from social.exceptions import AuthCanceled, AuthAlreadyAssociated, NotAllowedToDisconnect
 
 
 class TethysSocialAuthExceptionMiddleware(SocialAuthExceptionMiddleware):
@@ -39,6 +39,13 @@ class TethysSocialAuthExceptionMiddleware(SocialAuthExceptionMiddleware):
 
                 messages.success(request, blurb)
 
+                if request.user.is_anonymous():
+                    return redirect('accounts:login')
+                else:
+                    return redirect('user:settings', username=request.user.username)
+            elif isinstance(exception, NotAllowedToDisconnect):
+                blurb = 'Unable to disconnect from this social account.'
+                messages.success(request, blurb)
                 if request.user.is_anonymous():
                     return redirect('accounts:login')
                 else:


sdc50:
Drew,

You are exactly right. When a user is created with OAuth then that OAuth
connection is the only way to authenticate that user. If the OAuth account
were disconnected then the user would be orphaned and there would be no way
for that user to log in. If that user links another OAuth account (say
Google) then the original (HydroShare) account can be disconnected. It
would be ideal to have a way for the user to add a password, which I think
is in the plan. In the meantime Nathan made it so the disconnect button is
disabled if that is the only way to authenticate a user. That change is on
the latest version of dev and will be in the soon to be released 1.3.

zhiyuli:
@sdc50 @swainn 
I just checked out latest dev and confirmed the "disconnect button" will be disabled when the user has no tethys password or has only one associated social account. That is good.

I think this patch still has some value, it works as a final defense to prevent 500 error caused by the NotAllowedToDisconnect exception (if any). Thanks

swainn:
This looks good to me @zhiyuli. I don't think it will hurt to include this. +1

sdc50:
Thanks Drew! Did you end up getting your apps working again after upgrading?

zhiyuli:
Scott,

yes, latest dev is working on my end now.
I made the following changes in settings.py:

TEMPLATE_LOADERS = (
    'django.template.loaders.filesystem.Loader',
    'django.template.loaders.app_directories.Loader',
    # 'tethys_apps.utilities.tethys_apps_template_loader'
    'tethys_apps.template_loaders.TethysAppsTemplateLoader'
)

Maybe a more formal way is to re-create a new settings.py from scratch.

another issue after checking out dev is about the django-bootstraps3. The
template complained about failure to import some files related to
django-bootstraps3... Then I updated it to version 6.2.2 and that error was
gone.

Looks like 1.3 has many exciting major changes. Looking forward for an
official upgrade instruction.

Thanks
Drew

sdc50:
Drew,

I should have thought of this earlier when you were having problems, but
there are instructions for upgrading
<http://docs.tethysplatform.org/en/dev/installation/update.html> on the dev
version of the docs. We are just about to release 1.3 so the dev branch is
pretty much 1.3 right now.

zhiyuli:
Good to know the upgrade instruction is already online. Thanks.

swainn:
Yes. And I would point out that they have you regenerate your settings file. There are a lot more settings that have changed because of the upgrade from Django 1.7 to 1.9. Follow the instructions that Scott provided a link to noting that you will be using the dev branch instead of master until the official release.

PR #181, opened 2016-04-12
made selection possible for non wms layers

snowman2:
Added selection for non-WMS layers and added public init highcharts plot function for AJAX.


snowman2:
Added selection for non-WMS layers and added public init highcharts plot function for AJAX.

PR #182, opened 2016-04-25
Gizmo AJAX and Select Interaction for non-WMS layers

snowman2:
1. Added public function to plot view to enable gizmo ajax
2. Added public function to map view to enable gizmo ajax
3. Added select interaction to map view to enable selection for non-wms layers
4. Added image option for map view legend & a sub class for the geoserver legend image.


swainn:
I finally got to looking at this. Sorry for the delay. Looks good to me, but I'd like you to add documentation for the JavaScript endpoints you added for the plots and the map. Take a look at https://github.com/tethysplatform/tethys/blob/dev/docs/tethys_sdk/gizmos/map_view.rst for examples of what to do. Not sure if these features warrant a vote on the forum or not since it is just adding to existing features. What do you think @sdc50?

snowman2:
Documentation Added + Tethys bug fixes.

swainn:
This will be a great addition.

PR #183, opened 2016-05-11
Tags for apps

SarvaPulla:
A new feature for organizing apps. The developer will be prompted for tags during the setup. Once the tag are created, a list of buttons is dynamically generated to assist with the filtering (Courtesy of the Isotope JS Library).


swainn:
Sarva, 

I finally got around to reviewing your pull request. I really like this feature, but there were a few things that need fixing before I'd approve it including a major problem with scaffolding new apps.

Tweaks and Adjustments:

1. When no apps have tags, I'd like there to be no tag buttons displayed on the apps library page. Currently the "All" button shows up.
2. The prompt for tags during scaffold is a little confusing:  'Tags for filtering your apps. You can add multiple tags with spaces. e.g.: "Hydrology","Reference Timeseries"' Should read "with commas"???
3. The name of the template tag "get_class" seems a little too generic. Perhaps rename to "get_tag_classes" or "get_tag_classes_for_app" following the naming convention of your other template tag. Also there is some commented out code that needs to be removed.
4. I am unable to scaffold a new app. I always get this error: Error: Invalid characters in tags. No special characters are allowed. Only commas and spaces are allowed. 
5. Also, I should be able to scaffold an app without assigning any tags (i.e.: hit enter a bunch of times to accept all the defaults)

SarvaPulla:
Just pushed my latest code. Let me know if that's fine. Thanks!

PR #192, opened 2016-07-21
Created Delete tool and added Tooltips

astraiophos:
Hopefully I did things right this time, I feel confident about it though, everything worked on my local machine. Let me know if there's something wrong or something I should change


swainn:
Excellent. I like the tool tips a lot. There are a few things that needs to be done for this feature Described below. The first is to add documentation. For Gizmos, that involves two steps:

1) create a live demo of the gizmo on the Gizmo Showcase (in your case, modify the map view demo in Gizmo Showcase to have the Delete control enabled). The controller (called a "view" in Django lingo) for Gizmo showcase  is located at "tethys_gizmos/views/gizmo_showcase.py" and it is called "index". The MapView stuff starts at ~line 794.

There is also a separate page for showing of the map view by itself (http://localhost:8000/developer/gizmos/map-view) and it's controller is called "map_view" ~line 1095 of the gizmo_showcase.py. I'd add the control to this map as well. 

2) document the features in the doc strings of the appropriate Options classes. In this case, you need to add documentation to the MVDraw class which is located in "tethys_gizmos/gizmo_options/map_view.py" starting on ~line 250.

There are vector layers on both maps already, so you shouldn't have to do much other than add the Delete control to the maps I would imagine.

There is also a bug with the Delete feature:

The separate map view page (http://localhost:8000/developer/gizmos/map-view) demonstrates how the data is passed to the controller by displaying any drawn points on the screen after pressing the Submit button. It does so by updating a hidden text area field every time a change is made to the map. Use this page to make sure that the Delete control updates this text area appropriately. I did a quick test and it looks like it is not updating this field appropriately.

astraiophos:
For the hidden text area, do we care whether this textarea is updated dynamically or if it is updated with a specific trigger?

astraiophos:
How much does it matter if the text field has the following if all of the features are deleted from the collection:

`{"type":"GeometryCollection","geometries":[]}`

swainn:
Looks like the textarea is empty when the map first loads, so I don't think it is that important if it is empty when there are no other things in there.

swainn:
Is this ready for review again?

astraiophos:
It should be, i made changes and committed them, as long as i don't need to make another pull request or something
On Jul 29, 2016 10:18 AM, Nathan Swain <notifications@github.com> wrote:Is this ready for review again?

PR #199, opened 2016-12-02
Update hydroshare oauth backend

zhiyuli:
Currently most apps that work with hydroshare are hard coded to retrieve data from www.hydroshare.org.
There is a need for app work with different hydroshare instances, like beta.hydroshare.org.
This PR updated the hydroshare oauth backend to send more info to client (app controller), which could make app aware of which hydroshare instance user has logged in. So app can make following requests against that hydroshare instance. 

The changes have been manually applied on apps, appsdev and appsplayground servers. We would like to merge them into Tethys repo.


zhiyuli:
Once this gets merged, apps should make the following minor changes:
from tethys_services.backends.hs_restclient_helper import get_oauth_hs
hs = get_oauth_hs(request)
for resource in hs.getResourceList():
    print(resource)

tethys_services/backends/hs_restclient_helper.py
```diff
@@ -0,0 +1,29 @@
...
+from django.conf import settings
+
+
+def get_oauth_hs(request):
+    hs = None
+    error_msg_head = "Failed to initialize hs object: "
+
+    if request.user.social_auth.count() == 0:
+        raise Exception(error_msg_head + "Not logged in through OAuth")
+    if request.user.social_auth.count() > 1:
```
swainn: I'm pretty sure it is perfectly legal for users to link multiple social accounts to their Tethys account, in which case the current handling of multiple social auth objects would create a block to using the HydroShare client. Can you change this to support multiple social auth accounts?

tethys_services/backends/hydroshare_beta.py
```diff
@@ -0,0 +1,86 @@
...
+* Name: hydroshare.py
+* Author: Nathan Swain and Ezra Rice
+* Created On: July 31, 2015
+* Copyright: (c) Brigham Young University 2015
+* License: BSD 2-Clause
+********************************************************************************
+"""
+from social.backends.oauth import BaseOAuth2
+
+class HydroShareOAuth2(BaseOAuth2):
```
swainn: I think this class ought to be renamed HydroShareBetaOAuth2 and inherit from the HydroShareOAuth2 class, only overriding the necessary parameters. This would be a more DRY approach.

tethys_services/backends/hs_restclient_helper.py
zhiyuli: @swainn 
I was thinking about this. If user's tethys account gets associated with multiple social accounts, say hydroshare and hydroshare_beta..., which hydroshare instance should be used to retrieve data in following operations...

swainn: That's a good question. Ideally, the user is presented with a dialog and able to choose somehow. But in the case that that isn't an option, then you need to decide on a default behavior where one takes priority over the other. I don't really have a preference on that, so I'll let you decide if it comes to it.

zhiyuli:
@swainn 
I made following changes:
1) The get_oauth_hs() still works if user's tethys account gets associated with only one HydroShare social account and multiple other social accounts (like google or facebook). 
But if it is linked to multiple HydroShare social accounts (like hydroshare and hydroshare-beta), the get_oauth_hs() function raises an exception. We may change this later to let user choose from available hydroshare instances.

2) change class HydroShareBetaOAuth2 to inherit HydroShareOAuth2

swainn:
This looks good. I have two additional requests/requirements for approval: docs and tests. Please update the HydroShare entry in the  social authentication documentation if necessary (https://github.com/tethysplatform/tethys/blob/master/docs/tethys_portal/social_auth.rst#hydroshare). As of version 1.4.0, we will have a testing framework. Please write some tests for the HydroShare OAuth. @sdc50 can get you started.

zhiyuli:
@swainn 
will get this after AGU. thanks

zhiyuli:
@swainn 
I added some tests for hydroshare backend. please review. thanks

swainn:
@zhiyuli : This is looking great. One last request: will you please update the HydroShare social authentication documentation (https://github.com/tethysplatform/tethys/blob/dev/docs/tethys_portal/social_auth.rst) with the changes that you made? Also, add a blurb in the What's New page about the new features we are supporting related to your changes with a link to the documentation (https://github.com/tethysplatform/tethys/blob/dev/docs/whats_new.rst). See this page for examples of how the What's New page is laid out: https://github.com/tethysplatform/tethys/blob/dev/docs/whats_new/prior_releases.rst

PR #198, opened 2016-11-11
DataTables, Plotly, and Bokeh GIzmos + Reorder JS Loading + Conda YML

snowman2:
1. Added 3 new gizmos w/ documentation & examples:

- DataTablesView
- PlotlyView
- BokehView

2. Reordered JS so libraries load at top.
3. Added tag to load in JS libraries without gizmo. Good for gizmos AJAX
4. Added conda YML file (see: https://github.com/tethysplatform/tethys/issues/195).

PR #204, opened 2017-01-11
Erdc cm dev

sdc50:
This is essentially the same merge request from snowman2 but rebased and cleaned up after the conda_env merge request.


sdc50:
Replaces pull request #198

PR #216, opened 2017-01-30
Dev: A simple ESRI Map View GIzmo

SarvaPulla:
This gizmo uses the ESRI Javascript API Version 4.2. Create an ESRI Map Object with the EMView class. Create a view and add layers to that view using the EMView and the EMLayer classes.


swainn:
I'm excited for this. Can you please add an example in the Gizmo showcase for the ESRI map? See:
https://github.com/tethysplatform/tethys/blob/master/tethys_gizmos/templates/tethys_gizmos/gizmo_showcase/index.html

Also, documentation for it here: https://github.com/tethysplatform/tethys/tree/master/docs/tethys_sdk/gizmos

SarvaPulla:
@swainn Added Documentation and an Example in the gizmo showcase. It's good to go.

Issue #209, opened 2017-01-12
Brute force login attacks


snowman2:
Need ability to limit login attempts to prevent/deter brute force attacks.

snowman2:
Could be resolved by https://github.com/tethysplatform/tethys/pull/208

swainn:
@snowman2, was this resolved?

PR #227, opened 2017-02-27
Tokenauth

snowman2:
Modified to use API Key in authorization header for views specifically for API.


snowman2:
Example usage: 
https://github.com/erdc-cm/tethysapp-streamflow_prediction_tool/blob/msd/tethysapp/streamflow_prediction_tool/api.py

swainn:
@snowman2 I really have a lot of concerns with this approach, which @sdc50, @msouff, and I discussed during scrum last week, not the least of which is that the token never expires... This could represent a major security vulnerability. We need to develop a more robust, secure mechanism for API authentication. I think we ought to implement JSON Web Token based API authentication. I would also like this to be implemented as off-by-default with documentation describing how to turn it on. Check out these resources:

https://jwt.io/
http://getblimp.github.io/django-rest-framework-jwt/

sdc50:
I think there are two levels of tokens that could be used in Tethys. I think Alan is thinking the CKAN/Google API type tokens that give you access to a limited API.

The other type is a temporary token that would be used for a session after authenticating with user credentials. The problem with this type is that you can't allow for programatic login of the user or else you bypass any brute force attack fixes (see #209).

msouff:
I think that we will need both the static and the dynamic tokens. The reason is that for users logging in for various purposes a dynamic token makes sense. The way you get this token with JWT is by submitting a post request with your username and password. This would not work for an API, where we need the third-party app to be able to make calls without passing username and password as parameters or having them hard-coded. That would defeat the purposes of trying to be secured. A static token created specifically for that third-party app makes more sense in my opinion, which is what the current tokenauth does.

snowman2:
Also, it is good to note we will need to test this in production. See @msouff's comment.

https://github.com/erdc-cm/tethysapp-streamflow_prediction_tool/pull/5

snowman2:
Documentation added.
Be warned - made changes to the scaffold handoff.py file and added the esri map gizmo docs to a doctree. It was easier for me to just add those changes as I was going through the documentation.

msouff:
I actually have in production already. For this setup to work in production, SSL (https) needs to be enabled. This adds even more security when passing the token as a header.

PR #228, opened 2017-02-27
Feature/select2 options

sdc50:
Added the ability to pass select2 options to the select_gizmo.

### DIFF
--- a/tethys_gizmos/static/tethys_gizmos/js/select_input.js
+++ b/tethys_gizmos/static/tethys_gizmos/js/select_input.js
@@ -25,7 +25,11 @@ var TETHYS_SELECT_INPUT = (function() {
  	var initSelectInput;
 
  	initSelectInput = function(tag) {
-        $(tag).select2();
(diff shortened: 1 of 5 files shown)


tethys_gizmos/static/tethys_gizmos/js/select_input.js
```diff
@@ -25,7 +25,11 @@ var TETHYS_SELECT_INPUT = (function() {
  	var initSelectInput;
 
  	initSelectInput = function(tag) {
-        $(tag).select2();
+ 	    $(tag).each(function(){
```
swainn: Why the $.each here? initSelectInput is already being called once for each instance of the 'select2' class (see line 45ish). Would this be redundant?

sdc50: This is to make sure the functionality works correctly when being called from AJAX. Since I'm doing it here I could remove the .each from line 45ish. Thoughts?

Issue #145, opened 2015-10-02
Create an Installation Script


swainn:

sdc50:
With the move to conda I see two levels for this:

1. We package release versions as a conda package and then installing becomes trivial. A simple bash script could be used to set up an environment install the Tethys package and then walk the user through configuring the settings file.
2. A bash script to pull the development branch would also be possible. It would be especially easy if we set up the Tethys Database from conda rather than using the dockers.

swainn:
Perhaps we could install using a custom conda installer. No script required. Technically, for development installations, you don't really need to change any settings. We could even probably have it create a default super user.

PR #240, opened 2017-04-14
Install scripts

sdc50:
New install script that works on Linux and Mac. Stub for Windows install script also included. Updated documentation. Addresses issue #145.


scripts/install_tethys.bat
```diff
@@ -0,0 +1,61 @@
+# WARNING this is completely untested!!!!
```
swainn: How much work do you think this needs to get it going on Windows?

docs/installation/settings.rst
```diff
@@ -0,0 +1,7 @@
+***************
+Tethys Settings
```
swainn: While I like the idea of this... it really could just be a link to the Django settings page, no? It also occurs to me that we may have a name conflict with the Tethys "settings" as in settings.py and Tethys "settings" as in settings for an app... Could generate some confusion.

docs/installation/linux_and_mac.rst
```diff
@@ -0,0 +1,149 @@
...
+        * `--docker-options <OPTIONS>`:
+                Command line options to pass to the `tethys docker init` call if --install-docker is used. Default is "'-d'".
+        * `-x`:
+                Flag to turn on shell command echoing.
+        * `-h, --help`:
+                Print this help information.
+
+    Here is an example of calling the installation script with customized options::
+
+        $ . install_tethys.sh -t ~/tethys -a localhost -p 8005 -c ~/tethys/conda --db-username tethys_db_user --db-password db_user_pass --db-port 5437 -S tethys -E email@example.com -P tpass
```
swainn: I love how fully featured this is! I'm also glad that there are defaults for all of these, because that is a lot of options.

docs/installation/linux_and_mac.rst
```diff
@@ -0,0 +1,149 @@
...
+        * `-t, --tethys-home <PATH>`:
+                path for tethys home directory. Default is /usr/lib/tethys on Linux and /Library/Applicaiton/tethys on Mac OS.
+        * `-a, --allowed-host <HOST>`:
+                hostname or IP address on which to server tethys. Default is 127.0.0.1.
+        * `-p, --port <PORT>`:
+                port on which to server tethys. Default is 8000.
+        * `-b, --branch <BRANCH_NAME>`:
+                repository branch to checkout. Default is 'dev'.
+        * `-c, --conda-home <PATH>`:
+                path to conda home directory. Default is ${TETHYS_HOME}/miniconda.
```
swainn: Does miniconda need to be installed already for this to work? Or does this also install miniconda?

docs/installation/linux_and_mac.rst
```diff
@@ -0,0 +1,149 @@
...
+    You can customize your tethys installation by passing command line options to the installation script. The available options can be listed by running::
+
+         $ . install_tethys.sh --help
+
+    Each option is also descriped here:
+
+        * `-t, --tethys-home <PATH>`:
+                path for tethys home directory. Default is /usr/lib/tethys on Linux and /Library/Applicaiton/tethys on Mac OS.
+        * `-a, --allowed-host <HOST>`:
+                hostname or IP address on which to server tethys. Default is 127.0.0.1.
```
swainn: Typo: server => serve?

docs/installation/linux_and_mac.rst
```diff
@@ -0,0 +1,149 @@
...
+        * `-p, --port <PORT>`:
+                port on which to server tethys. Default is 8000.
```
swainn: Typo: server => serve?

docs/installation/linux_and_mac.rst
```diff
@@ -0,0 +1,149 @@
...
+        * `-b, --branch <BRANCH_NAME>`:
+                repository branch to checkout. Default is 'dev'.
```
swainn: Love this!

docs/installation/linux_and_mac.rst
```diff
@@ -0,0 +1,149 @@
...
+
+    (tethys) $ tethys manage start
+
+or simply just::
+
+    (tethys) $ tms
+
+.. tip::
+
+    The installation script adds several environmental variables and aliases to help make using Tethys easier. Most of them are active only while the tethys conda environment is activated, however one alias to activate the tethys conda environment was added to your `.bashrc` or `bash_profile` file in your home directory and should be available from the terminal:
```
swainn: change "available from the terminal" to "available from any terminal session"

docs/installation/update.rst
```diff
@@ -1,6 +1,6 @@
-****************
-Upgrade to 1.5.0
-****************
+********************
+Upgrade to |version|
```
swainn: I assume this is coming from pbr? Very nice. :D

environment_py2.yml
```diff
@@ -0,0 +1,45 @@
+# environment_py2.yml
```
swainn: is there a environment_py3.py in the works?

scripts/install_tethys.bat
sdc50: It's hard to say. Setting up the shell script took me a lot longer than I expected, but I kept adding features, and since shell scripting is not my forte it was a lot of Googling to try to figure out how to make it all work. Testing the script can take a while since it needs to download a lot of things, and since I was doing it on multiple platforms that made it even longer. All of that is to say I think it will be a fair about of work, but hopefully less that what has been done already.

sdc50: I should also say that I won't have time to work on this anymore for a while, so I was hoping you could pick it up from here.

docs/installation/settings.rst
sdc50: I think we have several settings that are Tethys specific (e.g. TETHYS_DATABASES), and also several that come from Django plugins. I think it would be helpful to list all of the settings in the docs with a brief description and have links to the Django/plugin documentation where it applies. 

As for the conflict with app settings, this is a good point. Maybe we should call this file platform_settings.py? I'm not sure how best to avoid the confusion. Maybe we just need to explain it clearly in the docs.

docs/installation/linux_and_mac.rst
sdc50: Thanks.

docs/installation/linux_and_mac.rst
sdc50: the `conda-home` directory is where miniconda will be installed.

docs/installation/update.rst
sdc50: Actually this is standard Sphinx! I used pbr to add a `|branch|` substitution so we can add that anywhere to dynamically set the branch anywhere in the docs.

environment_py2.yml
sdc50: the `environment_py2.yml` file is meant to replace the `tethys_conda_env.yml` file. Once we add Python 3 support for then we can create a `environment_py3.py` and make any necessary changes (which may only be changing the python version from 2 to 3). If we are dropping Python 2 support once we move to 3 then maybe we should just call this `environment.yml` and change it once we move to 3.

docs/installation/linux_and_mac.rst
sdc50: I wanted to make it super easy to install with defaults, but also wanted to provide flexibility to customize during the installation. One thing that I didn't get to, but think would be nice is to make it an interactive installer (like the Docker containers are) so that it will prompt the user for each value. You could still pass a flag (-d) to skip the interactive prompts and just use the defaults.

sdc50:
I just pushed some changes to fix the typos.

scripts/install_tethys.bat
```diff
@@ -0,0 +1,61 @@
...
+
+# create conda env and install Tethys
+conda env create -f tethys_conda_env.yml
+. activate tethys
+python setup.py develop
+tethys gen settings -d "%TETHYS_HOME%/src/tethys_apps"
+
+# Setup local database
+conda install --yes -c conda-forge postgresql
+initdb  -U postgres -D "%TETHYS_HOME%/psql/data"
```
sdc50: Here is a command to open a port on Windows that probably will be needed when someone get's to this:

```
netsh advfirewall firewall add rule name="postgresql" dir=in action=allow protocol=TCP localport=%TETHYS_DB_PORT%
```

PR #251, opened 2017-05-10
Comparing python2 to python3

SarvaPulla:
Ran the 2to3 module on the whole repository. This will give you an idea of what the code will look like in python3.


sdc50:
This was used as a reference for PR #263, which is now merged so this PR is no longer needed.

PR #255, opened 2017-05-11
Enhance Install scripts

sdc50:
made install scripts more robust
updated docs
cleaned up cli commands fixes

### DIFF
--- a/docs/installation/linux_and_mac.rst
+++ b/docs/installation/linux_and_mac.rst
@@ -33,7 +33,7 @@ For Systems with `curl` (e.g. Mac OSX and CentOS):
 
     You can customize your tethys installation by passing command line options to the installation script. The available options can be listed by running::
 
-         $ . install_tethys.sh --help
+         $ bash install_tethys.sh --help
 
     Each option is also descriped here:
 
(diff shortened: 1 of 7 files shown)


sdc50:
@swainn I realized that the reason you were always getting a `[process complete]` after running the script is because it always ended in an error. I have a setting to exit the script when an error is encountered, but since the script is being sourced (meaning it runs in the current shell) it ends up exiting the terminal. I've changed the docs to suggest running the script with `bash install_tethys.sh`. This will run in a sub shell so if there is an error it won't exit the terminal. This has the effect that the environment is not left active when the script ends (which is what you wanted anyway). I've updated the docs to reflect that.

As a side note, the error you were always getting was because the script was trying to check a variable that is only set on Linux (to install the dockers). I've fixed that so you shouldn't see that error anymore.

Also, I cleaned up the `manage_commands.py` to get rid of some of the variables that we weren't using anymore. I also added some code in `docker_commands.py` to handle Windows logging for the `docker init` command although that code is not tested.

sdc50:
@swainn also, a TETHYS_HOME path with spaces works on Windows as long as you specify a miniconda path that doesn't have spaces (because miniconda won't work if it's path has spaces). The Mac/Linux script does not yet work if the paths have spaces.

docs/installation/linux_and_mac.rst
```diff
@@ -86,12 +86,24 @@ For Systems with `curl` (e.g. Mac OSX and CentOS):
         $ . install_tethys.sh -t ~/Workspace/tethys -a localhost -p 8005 -c ~/miniconda3 --db-username tethys_db_user --db-password db_user_pass --db-port 5437 -S tethys -E email@example.com -P tpass
```
swainn: Should this example be changed to 'bash' as well?

sdc50: Yes, thank you I'll update that.

sdc50:
Just pushed some changes fixing the missing `bash` in couple of places in the docs. Also pushed some changes that allow the tethys home directory to contain spaces on Mac/Linux.

Issue #262, opened 2017-05-20
Alternative to pastescript


snowman2:
Pastescript has not been updated since 2015 and is not supported above 3.4. We need to find an actively developed alternative that supports python 2/3.

http://stackoverflow.com/questions/9356190/alternatives-to-python-pastescripts-paster-create

swainn:
@snowman2 I agree. I've been wanting to replace this using the native Django templating language, but it will take some doing. We should add this issue as a potential 2.1 priority.

PR #263, opened 2017-05-20
Python 2/3 compatible code

snowman2:
This needs to be tested, but it is a good start.

Note: This is only the source code for tethys and not the dependencies.

See:
https://github.com/tethysplatform/tethys/issues/262


snowman2:
Issue #242 needs to be fixed before the merge can happen.

snowman2:
Installation worked:
```
$ bash install_tethys.sh --python-version 3
```
NOTE: Installation will only work if you merge the code. Otherwise, tweaking needed.

environment_py3.yml
```diff
@@ -0,0 +1,46 @@
...
+- pyopenssl
+- pip:
+    - django-gravatar2
+    - django-bootstrap3
+    - django-model-utils
+    - django-guardian
+    - django-simple-captcha
+    - django-termsandconditions
+    - tethys_dataset_services
+    # - condorpy
```
sdc50: I'm just about to release condorpy so lets add this back in

scripts/install_tethys.sh
```diff
@@ -219,10 +225,20 @@ then
 
     # create conda env and install Tethys
     echo "Setting up the ${CONDA_ENV_NAME} environment..."
-    conda env create -n ${CONDA_ENV_NAME} -f environment_py2.yml
+    conda env create -n ${CONDA_ENV_NAME} -f "environment_py${PYTHON_VERSION}.yml"
     . activate ${CONDA_ENV_NAME}
     python setup.py develop
 
+    if [ $PYTHON_VERSION == '3' ]
```
sdc50: with condorpy released we can remove this section

Issue #258, opened 2017-05-15
Add persistent stores to the Web Admin Setup docs


sdc50:
The [Web Admin Setup](http://docs.tethysplatform.org/en/dev/installation/web_admin_setup.html) docs contain instructions for registering a 52 North server, a GeoServer, and a CKAN server. We probably should add instructions for adding a database server since that is now handled in the same way.

Issue #246, opened 2017-05-03
App Skin Change


swainn:
Update the app skin for 2.0. Ideas on this thread on the forum: https://groups.google.com/forum/#!topic/tethysplatform/YHfIkLzthA4

Issue #127, opened 2015-08-10
Add Update from 1.1 to 1.2 documentation


swainn:
Also add documentation for installing apps on a production server from Jiri.

Install a Tethys App on the Production Server
Step 1: Get the source code of the app
cd /usr/lib/tethys/src
sudo git clone https://github.com/my_name/tethysapp-my_first_app.git

(NOTE: instead of my_name and my_first_app, you put your name and the real name of your app)
 Step 2 Change the ownership of the files to the Apache user 
(on Ubuntu Linux the name of the Apache user is www-data. On Centos Linux the name of the Apache user is apache)
$ sudo chown –R www-data:www-data /usr/lib/tethys/src /var/www/tethys/static /var/www/.tethyscluster

Step 3 Run Installation Script of the App
The installation script (setup.py) will copy the app’s files to the correct directory, and it will install any dependencies that are required by the app.
   $ sudo su
   $ . /usr/lib/tethys/bin/activate
(tethys) $ cd /path/to/tethysapp-my_first_app
(tethys) $ python setup.py install
(tethys) $ exit

Step 4 Run Collectstatic
The static files of the app need to be collected into the directory that you created. Enter the following commands and enter “yes” if prompted.
   $ sudo su
   $ . /usr/lib/tethys/bin/activate
(tethys) $ tethys manage collectstatic
(tethys) $ exit

Step 5 Setup the Persistent Stores for the App
If your app uses a database, you need to run these commands to make and populate the app’s database
$ sudo su
   $ . /usr/lib/tethys/bin/activate
(tethys) $ tethys syncstores my_first_app
(tethys) $ exit

Step 6 Change Ownership again
(on Ubuntu Linux the name of the Apache user is www-data. On Centos Linux the name of the Apache user is apache)
sudo chown –R www-data:www-data /usr/lib/tethys/src /var/www/tethys/static /var/www/.tethyscluster

Step 7 Enable Site and Restart Apache
Finally, you need to disable the default apache site, enable the Tethys Portal site, and reload Apache:
On Ubuntu:
$ sudo a2dissite 000-default.conf && sudo a2ensite tethys-default.conf && sudo service apache2 reload

On Centos:
$ sudo service httpd restart

Tip 
You can quickly install an additional app using the following commands:
$ cd /usr/lib/tethys/src
$ sudo su
$ git clone https://github.com/my_name/tethysapp-my_second_app.git
$ . /usr/lib/tethys/bin/activate
[...]

PR #265, opened 2017-05-23
Tethys Skin Update

swainn:
Updated app skin for Tethys 2.0. Includes leaner header and app actions section, redesigned nav toggle and exit buttons, a new button in the header for quick access to settings, and a new block for adding custom buttons to the header. Pertinent documentation has been updated as well.

In addition the scaffold has been revamped to include demonstrations of proper usage of modals and the new header_buttons block. The app_base.css is minified and the header and app_actions section have been refactored to use flex box css styling. The random colors that new apps are generated with have been changed as well.

This pull request also includes commits that address #127, #258, and #246.


docs/whats_new.rst
```diff
@@ -12,9 +12,13 @@ Release |version|
 Powered by Miniconda Environment
 --------------------------------
 
-TODO
+* Tethys Platform is now installed in a Miniconda environment.
+* Using the Miniconda includes Conda, an open source Python package management system
+* Conda can be used to install Python dependencies as well as system dependencies
+* Installing packages like GDAL or NetCDF4 are as easy as ``conda install gdal``
```
snowman2: I would mention conda-forge is the recommended channel for installing packages.
```
conda install -c conda-forge gdal
```

sdc50: Should we just add conda-forge as the default channel in the install script? Either way some documentation about channels would be useful.

PR #269, opened 2017-06-01
tethys-264 MapView - ServerVector does not exist

swainn:
Removed ServerVector vector source and revamped all the supported sources. Need to come up with more robust support for ol.formats to more adequately support ol.source.Vector. Fixes 264.

### DIFF
--- a/tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
+++ b/tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
@@ -442,12 +442,13 @@ var TETHYS_MAP_VIEW = (function() {
     var GEOJSON = 'GeoJSON',
         KML = 'KML';
 
-    var TILE_SOURCES = ['TileDebug', 'TileImage', 'TileUTFGrid', 'Stamen', 'TileArcGISRest', 'Zoomify', 'XYZ', 'WMTS',
-                        'TileWMS'];
+    var TILE_SOURCES = ['TileDebug', 'TileUTFGrid', 'UrlTile', 'TileImage', 'VectorTile', 'BingMaps', 'TileArcGISRest',
+                        'TileJSON', 'TileWMS', 'WMTS', 'XYZ', 'Zoomify', 'CartoDB', 'OSM', 'Stamen'];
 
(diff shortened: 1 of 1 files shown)


snowman2:
It looks good. I think it can be merged. But, I think it would be a good idea to leave the issue open until we get a replacement for the functionality of ServerVector. Unless we want to move away from that and only support the WMS-WFS combo.

swainn:
@snowman2 what is the functionality of ServerVector? I don't know that I've used it before. If it is gone, it is gone. Not much we can do about it. With this commit I have added "support" for every Vector, Image, and Tile ol.source class there is in version 4, but only the ones that are documented in MapView have been tested. It looks to me that many of the Vector sources require the use of an ol.format class, which is essentially a text parser (think GeoJSON, GML, KML), and to date we only support GeoJSON and KML ol.format. Issue #270 will address that, but probably not until 2.1 at the earliest.

snowman2:
@swainn, it supports vector WFS. An example of the current method of doing the same thing is here: http://openlayers.org/en/latest/examples/vector-wfs.html?mode=advanced&q=WFS

snowman2:
```javascript
      var vectorSource = new ol.source.Vector({
        format: new ol.format.GeoJSON(),
        url: function(extent) {
          return 'https://ahocevar.com/geoserver/wfs?service=WFS&' +
              'version=1.1.0&request=GetFeature&typename=osm:water_areas&' +
              'outputFormat=application/json&srsname=EPSG:3857&' +
              'bbox=' + extent.join(',') + ',EPSG:3857';
        },
        strategy: ol.loadingstrategy.bbox
      });
```

swainn:
@snowman2 Ah. I see. The ol.source.Vector is "supported", but that particular example probably wouldn't work because of the function value of url...  I believe in the HighCharts interface we had at one point allowed passing JavaScript functions as strings and I had something on the JavaScript site that could convert it to a function. So, the pieces are there to support that kind of thing. In fact, I think at some point @sdc50 and I had discussed making our Python interface to OpenLayers more comprehensive. The other challenge with that is the "new" classing all over the place with the open layers API. I'll go ahead and merge this for now, but we ought to discuss this more in the future.

Issue #261, opened 2017-05-17
tethys docker intit -c postgis geoserver (Message)


snowman2:
``` 
$ tethys docker intit -c postgis geoserver
```
It said:
```
52 North WPS Docker container already installed: skipping.
```
Should say:
```
52 North WPS Docker container: skipping.
```

PR #272, opened 2017-06-02
tethys-261 tethys docker intit -c postgis geoserver 

swainn:
Fix for issue #261. Removed conflicting "already installed" messages.


snowman2:
@swainn, it looks good. I tested it and it works!

PR #280, opened 2017-06-05
added enable to script so it starts up again when computer restarts

snowman2:


snowman2:
@sdc50, I tested on RHEL and rebooted the machine and docker was not running. This should fix that.

PR #284, opened 2017-06-11
Tethys 256 scaffolding

swainn:
Implemented new scaffolding command that does not depend on paster/paste script. The command should be cross-platform.


swainn:
@sdc50 can you review and merge as soon as possible so that @SarvaPulla can test in Windows?

swainn:
@sdc50 the startapp solution didnt give me quite enough control to do what we needed scaffolding to do. This didn't take too long to roll and I think it will serve us better.

PR #288, opened 2017-06-17
Bugfix: Postgis extension enabling now functions

shawncrawley:
The problem was that the connection on which the enable postgis extension command must be run is the connection to the database itself. I simply passed the `with_db=True` param to the `get_engine` call.


swainn:
@shawncrawley thanks for catching and patching!

Issue #281, opened 2017-06-08
Python version option missing from installation script docs


swainn:
Need to update docs to show Python version option.

Issue #271, opened 2017-06-01
Tethys Dev Install Guide


mmbayles:
On the install page[ instructions for linux and mac](http://docs.tethysplatform.org/en/dev/installation/linux_and_mac.html) the default sign credentials are not easily found in the instructions. I would recommend stating the credentials in the main instruction area.

sdc50:
@mmbayles thanks for pointing this out. I've added the default credentials in the documentation more explicitly in a PR that should be merged soon.

Issue #243, opened 2017-04-21
Squash merge migrations or regenerate migrations


swainn:
Scott's approach...

swainn:
@sdc50 hold off on this for a bit. I'm upgrading to Django 1.11 and there were a few migrations to apply.

swainn:
@sdc50 OK. This is good to go after merging #275

Issue #146, opened 2015-10-02
Create CentOS 7 installation instructions.


swainn:
See the following issues: #137, #136, #64 for reference. 

We need to find a solution that does not involve disabling SELinux. Here on some notes on SELinux to get you started:

http://blog.endpoint.com/2010/02/selinux-httpd-modwsgi-26-rhel-centos-5.html
https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/4/html/SELinux_Guide/selg-preface-0011.html

PR #291, opened 2017-06-19
Production installation

sdc50:
Working production installation with squashed migrations.


sdc50:
Fixes #281
Fixes #271
Fixes #146
Fixes #60
Fixes #243

PR #308, opened 2017-07-06
Geoserver Database Model

snowman2:


swainn:
@sdc50 and I recommend implementing this as an extension. If it gains wide adoption, then we could consider merging it into core.

snowman2:
This is pretty old and likely outdated - and I completely forgot about it. I think closing it is appropriate.

sdc50:
@snowman2 I know that you are busy with other things (like completely rewriting pyproj 🥇), so we should have mentioned that I was planning to take on updating this PR and converting it into an extension. We like the idea, but just haven't gotten to working it in yet. I'm going to reopen to make sure I don't forget about it.

snowman2:
Sounds good 👍

PR #312, opened 2017-07-20
Bug fixes to the testing framework

shawncrawley:
In my attempts to finish migrating my app to Tethys 2.0, I came across a number of bugs with the testing framework. Though this commit contains the fixes to these bugs, there were a couple hokey things that I had to do to achieve the fixes. This makes me wonder if certain things need to be addressed, such as:
*See my added comments*


shawncrawley:
When creating tethys-testing_ copies of databases, I jimmy-rigged the create_persistent_store function to be able to be called with no connection_name only when testing, because it will use the same connection as the existing database. I thought of creating a new function called create_test_persistent_store, but I figured that it would just duplicate 90% of the code, so I just tweaked the existing function instead...

Issue #249, opened 2017-05-09
Handle unset settings for newly installed apps


sdc50:
Now that we've moved to having app settings it is possible for app developers or portal administrators to shoot themselves in the foot with having unset settings. If an app tries to import a setting in the global space of a module then that can cause an error that prevents Tethys from starting when that app is installed, and thus the setting has no way of being set. 

Also if an administrator installs an app but forgets to set the settings then the app will likely throw an error. It would be great to have warning messages printed when the server starts indicating that there are required settings that need to be set.  Also, having the app redirect to the settings page rather than allowing it to throw an error when settings need to be set would be a nice feature.

If possible it would be great for the app to not even be publicly visible until it has been properly configured by the admin.

swainn:
Agreed. We definitely need to catch the settings errors. I like the idea of redirecting to the settings page. We have the module level issue with anything that tries to use a Django model on startup of Tethys. I think we could probably just wrap our model imports in try, except statements and just log a warning when it fails. Then we won't have that constraint on startup. We just need to be careful not to mask the error.

sdc50:
@swainn I see this as a partial fix. It will at least prevent the problem of getting the server "stuck" where it won't start because of an error in an app, but also won't allow you to uninstall the app. I'm about to issue another PR to master that is the second phase of a more complete fix.

PR #316, opened 2017-08-17
Hotfix/catch settings errors

sdc50:
Prints warning rather than raising an error when an unset setting is accessed during app loading.

Partially fixes issue #249.


sdc50:
This completely fixed the wrong thing. Replaced by PR #318

PR #317, opened 2017-08-17
Refactor harvesting

sdc50:
1st (includes commits from PR #316) and 2nd phase in fixing issue #249. This reorders the way that apps are loaded and is able to catch more errors on loading and prevents apps from being loaded that have errors while not stopping the whole server from starting.

### DIFF
--- a/tethys_apps/__init__.py
+++ b/tethys_apps/__init__.py
@@ -7,20 +7,10 @@
 * License: BSD 2-Clause
 ********************************************************************************
 """
-import logging
-import warnings
 
 # Load the custom app config
 default_app_config = 'tethys_apps.apps.TethysAppsConfig'
 
-# Configure logging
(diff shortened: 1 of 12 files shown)


sdc50:
The result of some apps not being "loaded" is that the database does not always have all of the apps listed by `get_installed_tethys_apps`. I've ensured that this works with the `uninstall` command, but there are three other uses of `get_installed_tethys_apps` that we need to decide what the appropriate behavior should be.

sdc50:
Maybe we need to add a field to the apps database that stores what the status of the app is (i.e. uninitialized, initialized, error). That way all installed apps are in the DB, but we can still separate the apps that have an error and not try to load them in.

sdc50:
@swainn Don't check this yet. I just found a bug. I'll push another commit sometime today.

sdc50:
replaced by #319

PR #319, opened 2017-08-21
App loading refactor

sdc50:
Refactor of app settings and app harvesting to make it more robust (i.e. it won't fail if one app has an error while loading). Also fully handles unconfigured apps in the GUI: only admins will see the apps in the Apps Library, but they will be styled differently and when clicked will go directly to the app settings page (see below).

image

There are a few impacts that still need to be considered (all marked with TODO comments in the code).


swainn:
Merge into Master branch for further development.

PR #320, opened 2017-09-20
Exception handling middleware (AJAX)

snowman2:


snowman2:
Hi @sdc50, I have examples [here](https://github.com/erdc/tethysapp-streamflow_prediction_tool/blob/master/tethysapp/streamflow_prediction_tool/controllers_ajax.py). Just replace the local imports with the tethys imports.

snowman2:
Also, for the JS side, [this](https://github.com/erdc/tethysapp-streamflow_prediction_tool/blob/master/tethysapp/streamflow_prediction_tool/public/js/main.js#L306-L332) is helpful.

swainn:
@sdc50 and I recommend implementing this as an extension. If it gains wide adoption, then we could consider merging it into core.

snowman2:
This is pretty old and likely outdated - and I completely forgot about it. I think closing it is appropriate.

sdc50:
@snowman2 I'll take this one too (eventually).

snowman2:
Sounds good 👍

PR #338, opened 2018-04-14
Python3 compatibility

sdc50:
changes to migrations and code to ensure python2/3 compatibility.


sdc50:
@swainn when I tried to migrate the database with the`0004_tethysextension.py` it would always generate another migration. I ended up editing the migration to include everything and renamed it to be `0002_tethysextension.py`. I think you should be able to migrate your databases to the new migration by adding the following to `0002_tethysextension.py`:

```python
class Migration(migrations.Migration):
    replaces = [('tethys_apps', '0004_tethysextension')
    ...
```

Add that line in and then run migrations. Then you should be able to delete `0004_tethysextension.py`and remove the line in `0002_tethysextension.py`.

PR #345, opened 2018-08-29
PEP8

sdc50:
Lots of minor changes to conform with PEP8 style guides. (needs to be rebased)


sdc50:
@swainn I would suggest that someone checkout this branch and rebase it onto master and then submit it as a new PR.

swainn:
Will get merged with Python 3 PR that is forthcoming.

PR #346, opened 2018-08-31
jobs refactor

sdc50:
* deprecate job templates
* add ability to create job without template
* fix jobs table bugs
* add ability to dynamically add the id as part of the job name

### DIFF
--- a/docs/tethys_sdk/jobs.rst
+++ b/docs/tethys_sdk/jobs.rst
@@ -39,7 +39,7 @@ To create jobs in an app you first need to define job templates. A job template 
                                                 remote_input_files=('$(APP_WORKSPACE)/my_script.py', '$(APP_WORKSPACE)/input_1', '$(USER_WORKSPACE)/input_2'),
                                                 executable='my_script.py',
                                                 transfer_input_files=('../input_1', '../input_2'),
-                                                transfer_output_files=('example_output1', example_output2),
+                                                transfer_output_files=('example_output1', 'example_output2'),
                                                 )
 
       job_templates = (CondorJobTemplate(name='example',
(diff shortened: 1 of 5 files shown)


sdc50:
Here is an example of how the jobs api will change:

### The old way
In `app.py`:

```python
    def job_templates(cls):
        """
        Example job_templates method.
        """

        my_job_description = CondorJobDescription(condorpy_template_name='vanilla_transfer_files',
                                                  remote_input_files=('$(APP_WORKSPACE)/my_script.py', '$(APP_WORKSPACE)/input_1', '$(USER_WORKSPACE)/input_2'),
                                                  executable='my_script.py',
                                                  transfer_input_files=('../input_1', '../input_2'),
                                                  transfer_output_files=('example_output1', 'example_output2'),
                                                  arguments=['arg1', 'arg2']
                                                  )

        job_templates = (CondorJobTemplate(name='example',
                                           job_description=my_job_description,
                                           scheduler=None,
                                          ),
                        )

        return job_templates
```

In `controllers.py`:
```python
    job = job_manager.create_job('template_job', request.user, 'example')
    job.save()
```
### The new way
The `job_templates` method on the app class is now deprecated. All jobs are instantiated and defined in a controller (or helper script). 

In `controllers.py`:
```python
    app_workspace = app.get_app_workspace().path
    job = job_manager.create_job(
        name='job_{id}',
        user=request.user,
        job_type='CONDOR',
        attributes=dict(attribute1='attr1'),
        condorpy_template_name='vanilla_transfer_files',
        remote_input_files=(
            os.path.join(app_workspace, 'my_script.py'),
            os.path.join(app_workspace, 'input_1'),
            os.path.join(app_workspace, 'input_2')
        )
    )
    # attributes and properties can also be added after the job is created
    job.set_attribute('executable', 'python.py')
    job.extended_properties = {'one': 1, 'two': 2}
    job.save()
```

swainn:
Will be included in forthcoming Dask Job Type PR.

PR #349, opened 2018-11-01
Input fix

gagelarsen:
This was a fix for compatibility issues between python 2 and python 3. The scaffold command was using the 'input' function which isn't native to python 2. That is fixed by importing input from builtins.


swainn:
@SarvaPulla and @sdc50 this is the fix for the scaffolding issue in Python 2 of the new branch. Please review and merge if acceptable.

PR #352, opened 2018-11-14
add new tests with outdated removed

zhiyuli:
new intermediate tests added with outdated ones removed


swainn:
New tests similar to these were added with @msouff django 2.1 PR: #409

PR #357, opened 2018-11-27
Feature/basemap options

sdc50:
Provides more options for base map services and allows the OpenLayers version to be specified.


tethys_gizmos/gizmo_options/map_view.py
```diff
@@ -219,6 +249,35 @@ class MapView(TethysGizmoOptions):
...
+        ]
+        esri_layers = [{'ESRI': {'layer': l}} for l in esri_layer_names]
+        basemaps = [
+            'Stamen',
+            {'Stamen': {'layer': 'toner', 'label': 'Black and White'}},
+            {'Stamen': {'layer': 'watercolor'}},
+            'OpenStreetMap',
+            'CartoDB',
+            {'CartoDB': {'style': 'dark'}},
+            {'CartoDB': {'style': 'light', 'labels': False, 'label': 'CartoDB-light-no-labels'}},
```
swainn: It’s somewhat confusing to have both the label and labels options. Perhaps we rename label to control_label or similar.

tethys_gizmos/gizmo_options/map_view.py
```diff
@@ -266,16 +329,27 @@ def __init__(self, height='100%', width='100%', basemap='OpenStreetMap', view={'
         self.disable_basemap = disable_basemap
         self.feature_selection = feature_selection
 
-    @staticmethod
-    def get_vendor_js():
+    @classmethod
+    def static_url(cls):
```
swainn: As we discussed, let’s handle the in between cases CDN >= 5.0.0 in case they support those versions in the future.

tethys_gizmos/gizmo_options/map_view.py
```diff
@@ -266,16 +329,27 @@ def __init__(self, height='100%', width='100%', basemap='OpenStreetMap', view={'
...
+        return cls.cdn if cls.ol_version != '5.3.0' else cls.local_url
+
+    @classmethod
+    def debug(cls):
```
swainn: Whoops. Guess the previous comment should be here.

PR #358, opened 2018-11-30
Deprecate Python 2

### LINKED ISSUE #267: Upgrade Plotly
This is currently pinned at a lower version. Need to investigate why...

### LINKED ISSUE #309: Tethys Install Script Should Be Able to Configure an Existing Database (not Conda) for Production
Installing Tethys in production using the install script is only able to setup the Conda PostgreSQL server that is installed with Tethys and cannot be configured to setup a different PostgreSQL. To have Tethys use a different PostgreSQL instance, it is necessary to manually change the settings and initialize the database, which is not ideal.

This is problematic, because the Conda PostgreSQL may not be suitable for production. Also, admins may want to setup the  database server on a separate machine from the start.

### LINKED ISSUE #353: Depreciation warning for Python 2
Print warning if Tethys is running in Python 2 environment to indicate that support will be dropped in Tethys 3.0.

sdc50:
fixes #353
fixes #309
fixes #267


sdc50:
@swainn this is ready to go now.

Issue #285, opened 2017-06-12
Update Admin Pages documentation


sdc50:
It looks like the admin pages docs are out of date. In particular I think we need to add a section on app settings.

PR #369, opened 2018-12-13
Geoserver port changes & Admin Docs Update

swainn:
This PR addresses issues #277 and #285. 

Also, fixed an issue with the Tethys uninstall command that would occur if there was an app in the the database without a corresponding module installed. In this case, you could not uninstall the app in the database and the apps library page would show an error. There is likely a problem with the db sync that is allowing this to happen that needs to be addressed in the future.


sdc50:
@SarvaPulla Are you good if I merge this?

swainn:
@SarvaPulla go ahead and merge if you are the last one to review!

PR #375, opened 2018-12-18
Whats new & Upgrade Documentation

swainn:
- Updated What's New content
- Updated Upgrade Instructions


swainn:
Note: the upgrade instructions are preliminary and awaiting Scott's PR with changes to install script to facilitate using it for updating existing tethys installations.

swainn:
@sdc50 I updated per your comments on the new install script. The one thing that I don’t love is that the version is hard coded in the docs as part of the environment names. I tried to use PBR to handle that but it wasn’t working out. Take a look.

docs/installation/update.rst
```diff
@@ -2,42 +2,23 @@
...
-Install new dependencies and upgrade old ones:
+1. Activate Tethys environment and start your Tethys Database:
 
 ::
 
-             $ . /usr/lib/tethys/bin/activate
-    (tethys) $ pip install --upgrade -r /usr/lib/tethys/src/requirements.txt
-    (tethys) $ python /usr/lib/tethys/src/setup.py develop
+    . t
+    tstartdb
```
sdc50:
```suggestion
    tstartdb  # If you are using the locally installed database server
    # or
    tethys docker start -c postgis  # If you are using the docker container for your database server
```

sdc50: @swainn Let me know what you think about this update. I'm fine leaving it the way it is, but this may help to highlight that the `tstartdb` command won't start up everyone's database server.

PR #380, opened 2018-12-28
Release Candidate for 2.1.0

SarvaPulla:


swainn:
@sdc50 and @SarvaPulla sorry for the delay. I'm just getting back from holiday break and I didn't do a good job of watching for these updates. The only thing I'm concerned about at this time is that we haven't had a successful build of the "latest" docs for a while: https://readthedocs.org/projects/tethys-platform/builds/8312796/ I'm going to look into it.

sdc50:
It looks like it keeps failing because of a timeout. I think RTD has restricted their build times recently. For another project I had success mocking all of my dependencies to make the environment build faster (see https://docs.readthedocs.io/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules)

swainn:
> mocking all of my dependencies to make the environment build faster

Did you also not have it install the dependencies?

swainn:
@SarvaPulla, @sdc50 I just submitted a PR with the requirements mocked in docs builds. See #381

sdc50:
It looks like we are good to go. @swainn Do you want to pull the trigger on this?

swainn:
Looks good to me. Sure thing.

PR #381, opened 2019-01-02
Changes to speed up docs build on RTD and prevent timeouts

swainn:
* mock dependencies.
* install minimum required dependencies for docs build.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -54,9 +54,6 @@ print('NOTE: The following modules are mocked to prevent timeouts during the doc
 print('{}'.format(', '.join(MOCK_MODULES)))
 sys.modules.update((mod_name, MockModule()) for mod_name in MOCK_MODULES)
 
-# Patch mock the harvester harvest method
-mock.patch('tethys_apps.harvester.SingletonHarvester.harvest')
-
 # Fixes django settings module problem
 sys.path.insert(0, os.path.abspath('..'))
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,7 +14,6 @@
 
 import sys
 import os
-import re
 import mock
 
 import pbr.version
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -14,23 +14,65 @@
...
+    'django.contrib.contenttypes',
+    'django.contrib.sessions',
+    'django.contrib.messages',
+    'django.contrib.staticfiles',
+    'tethys_config',
+    'tethys_apps',
+    'tethys_gizmos',
+    'tethys_services',
+    'tethys_compute',
+]
```
sdc50: Why hard code this instead of parsing it out of the template like we did previously?

swainn: I don't want to load all of the apps. Basically the minimum we need are the built-in apps and the Tethys apps. The others cause problems, since we aren't actually installing them. 

Ideally, we would mock out Django and we wouldn't need any of these or the django.configure() and django.setup() lines. However, I wasn't able to quickly find a way to mock out Django entirely, because of some issues with mocking the Django model classes. So, these are the minimum required apps we need with this approach. We could probably refine this more for 3.0, but I think this will work for now.

swainn:
@sdc50 @SarvaPulla I've only tested this in my dev environment. I used new environments to ensure the docs build without installing all of the Tethys dependencies. Let me push the branch I'm using to origin so we can test building the docs before merging.

swainn:
> Let me push the branch I'm using to origin so we can test building the docs before merging.

It's building now: https://readthedocs.org/projects/tethys-platform/builds/8345968/

swainn:
@sdc50 @SarvaPulla The other question I have is that right now I'm still using the conda environment for the docs build, just installing minimum dependencies. We could probably speed up the build even more by dropping conda for docs builds and using pip. What do you think?

swainn:
btw, docs build succeeded: [RTD 8345968](https://readthedocs.org/projects/tethys-platform/builds/8345968/)

sdc50:
> btw, docs build succeeded: [RTD 8345968](https://readthedocs.org/projects/tethys-platform/builds/8345968/) And it completed in 6 minutes, so I don't think we need to worry about it taking too long anymore.

Great! I think just using pip (rather than conda) is a good idea, but maybe not worth the effort since it's building just fine now. 

I approve merging this now, but we still need to make sure to submit a PR to #380 as well.

swainn:
Ok. I’ll merge into master. @SarvaPulla van you then merge master into the branch associated with #380 ?

PR #387, opened 2019-01-29
Fix docker commands for python 3

tylorbayer:
Fix docker command tests for python 3

Issue with tethys docker command (issue with strings).
Added .decode() to fix.


swainn:
Just merged a new PR that addresses these and other issues with Docker. So, this PR is no longer needed.

Issue #384, opened 2019-01-09
Tethys Compute job status updates are broken in Master30


Tsubashi:
Jobs don't update their status because the `is_time_to_update` function checks first for the `_last_status_update` attribute which isn't assigned until `is_time_to_update` succeeds.
https://github.com/tethysplatform/tethys/blob/0b844d384d719486ce4a5171e67ecda67d2ddb6a/tethys_compute/models/tethys_job.py#L149-L150

To solve this I would change the existing `last_status_update` property to be lazily self-initialized and update `is_time_to_update` use the property, removing the need for the check altogether.

PR #400, opened 2019-02-25
Jobs table fix

swainn:
The jobs table was not updating as it should likely as a result of changes made during the development of the Dask job. The behavior of jobs that were Pending seems to have changed or been fixed. The primary fix in the PR is changing statuses from Pending to Submitted when a job is executed programatically.

### DIFF
--- a/tethys_gizmos/static/tethys_gizmos/js/jobs_table.js
+++ b/tethys_gizmos/static/tethys_gizmos/js/jobs_table.js
@@ -4,7 +4,6 @@
  *   copied from (https://docs.djangoproject.com/en/1.7/ref/contrib/csrf/)
  *
  *****************************************************************************/
-
 function getCookie(name) {
     var cookieValue = null;
     if (document.cookie && document.cookie != '') {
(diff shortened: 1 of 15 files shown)


swainn:
This addresses #384

swainn:
Addresses #405

swainn:
@SarvaPulla and @sdc50 this should be good to go. Included in this pull request:
* Addressed several issues that were causing the jobs table not to update including
    * PEN -> SUB when using execute() method
    * Use django timezones for datetimes in jobs models
* Addressed instabilities with the Dask Jobs API
    * Submitting many Dasks jobs would cause Tethys to lock up or crash
    * Dask distributed clients were not being closed properly
* Cleans up the Jobs table for Dask jobs showing only one instance of the Dask Task status at the top of the table instead of repeating it for each job.
* Mocked out new dependencies in the docs conf.py (we don't install the dependencies for the docs builds)
* Fixed a few hold-over docs typos for spatial dataset services.

tethys_gizmos/static/tethys_gizmos/js/jobs_table.js
```diff
@@ -204,14 +203,12 @@ function update_row(table_elem){
             });
             status = json.status;
             if(status == 'Running' || status == 'Submitted' || status == 'Various'){
-                var run_time_field = $('#run_time-' + job_id)
+                active_counter++;
```
sdc50: Is this being used?

sdc50: Got it. Looks like it's a global that gets initialized below.

sdc50:
@SarvaPulla the unofficial "official" policy of Tethys platform is that a PR can be merged as long as it gets Two approvals from the steering committee and no disapprovals. We haven't been great about following those guidelines, but maybe we should start. Since @swainn submitted the PR I think that counts as an implicit approval. I've just reviewed it and approve. Unless you have any objections then I think we should be good to merge this in.

SarvaPulla:
I don't have any objections. Feel free to merge it.

Issue #406, opened 2019-03-08
Advanced tutorial persistent stores really needs better documentation


waderoberts123:
Right now the steps are lacking in detail and very unfriendly for new users. Steps really should be made more clear on how to sync persistent stores. Even with more experience with Tethys it is hard to understand the current tutorials because they are very vague. I am currently running into this issue: 
<a href="http://tinypic.com?ref=sqtmok" target="_blank"><img src="http://i65.tinypic.com/sqtmok.png" border="0" alt="Image and video hosting by TinyPic"></a>
I made the new Persistent Store Service based on tutorial step 1g (Not sure what to name it because it wasn't specified)
<a href="http://tinypic.com?ref=2mm7dy0" target="_blank"><img src="http://i67.tinypic.com/2mm7dy0.png" border="0" alt="Image and video hosting by TinyPic"></a>
I added the new Persistent Store service to my app.
<a href="http://tinypic.com?ref=nwh6pz" target="_blank"><img src="http://i65.tinypic.com/nwh6pz.png" border="0" alt="Image and video hosting by TinyPic"></a>
I get this error.
<br>
So maybe it would be helpful to me if I had more experience with SQalchemy, but it seems as though this should be documented. When I am able to resolve this error I can update the docs and submit a PR.

waderoberts123:
I resolved the issue I was having and updated the documentation to be more clear. PR has been submitted.

swainn:
Thank you for your feedback @waderoberts123 , though the image of the error in your post is not showing up for me. I've added this to our milestone for the next release of Tethys (3.0). I'm going to reopen this until we have merged the PR to make sure it happens. I'll add a comment in the PR to link it to this issues.

waderoberts123:
The error was saying: "Role admin does not exist" since I was trying to use the "admin", "pass" instead of the database superuser credentials.

PR #407, opened 2019-03-08
Updates to the advanced tutorial to make more clear

waderoberts123:
I just added some minor changes to the tutorial so that future users will hopefully avoid any confusion and be able to utilize the persistent stores feature more easily.


swainn:
Addresses #406

PR #410, opened 2019-03-15
Cesium Support for Globe, Clock, and Constants

swainn:
Adds support on the Python end for:

- Globe options
- Clock options
- Use of Cesium constants

Also, fixes an issue with images in documentation that was introduced w/ PR #407


swainn:
nvm

PR #412, opened 2019-04-08
New MapView Features

swainn:
This Pull Request adds these new features to the MapView gizmo:
* Initial "drawn" features can be passed to the drawing layer when drawing tools are enabled.
* Snapping to a displayed layer is now possible when drawing features. The snapping layer can be either a vector layer or a WMS layer.
* Full support for specifying styles on vector layers in MVLayer
* Recursive, OL object parser from Python/JSON


tethys_gizmos/static/tethys_gizmos/css/tethys_map_view.min.css
```diff
@@ -1 +1,244 @@
...
\ No newline at end of file
+/*****************************************************************************
```
sdc50: this doesn't look minified.

environment.yml
```diff
@@ -6,8 +6,8 @@
 name: tethys
 
 channels:
-- tethysplatform
 - conda-forge
+- tethysplatform
```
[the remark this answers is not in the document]
swainn: @sdc50 this has been addressed and the file subsequently no longer shows up as changed.

tethys_gizmos/static/tethys_gizmos/css/tethys_map_view.min.css
swainn: @sdc50 tethys_map_view.min.css is now compressed.

PR #424, opened 2019-05-06
Migrations cleanup

sdc50:
Delete's old migrations. Should be rebased and merged after release of 3.0


swainn:
We decided to start from scratch on this.

PR #432, opened 2019-06-17
Fix for Issue 431 and Tweaks to Gen Command

swainn:
* Handles data migration for migration added in #423.
* Removes makemigrations from syncdb command
* Adds tethys_port to optional argument for gen command, instead of getting it from environment.


swainn:
Fixes #432

PR #440, opened 2019-07-03
Added Proxy app support

rfun:
This commit adds support to list apps in the settings page as Proxy apps, which will display them on the main app library page, but will basically redirect to the url entered. This is one step towards having split portals across servers and is also helpful during transition from Py 2 to 3 as we still have some apps that might be working only on the old python so far and hence we aren't ready to move them over yet. 

No documentation for this has been written since I don't see this being used by the general populous yet, but maybe in the future?


tethys_config/admin.py
```diff
@@ -37,4 +37,5 @@ def has_add_permission(self, request):
         return False
 
 
+admin.site.register(AppProxyInstance)
```
swainn: Move this to tethys_apps/admin.py

tethys_config/models.py
```diff
@@ -41,3 +42,21 @@ def as_dict(cls):
             settings_dict[code_name] = setting.content
 
         return settings_dict
+
+
+class AppProxyInstance(models.Model):
```
swainn: Move this to tethys_apps/models.py

tethys_config/models.py
```diff
@@ -41,3 +42,21 @@ def as_dict(cls):
...
+    """
+
+    name = models.CharField(max_length=100, unique=True)
+    endpoint = models.CharField(max_length=1024, validators=[validate_url])
+    logo_url = models.CharField(max_length=100, validators=[validate_url], blank=True)
+    description = models.TextField(max_length=2048, blank=True)
+
+    class Meta:
+        verbose_name = 'App Proxy Instance'
+        verbose_name_plural = 'App Proxy Instances'
```
swainn: Change to "Proxy App" and "Proxy Apps", respectively.

tethys_config/models.py
swainn: Change to ProxyApp

tethys_config/admin.py
swainn: Move to tethys_apps/admin.py

tethys_config/migrations/0002_appproxyinstance.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.20 on 2019-07-03 01:45
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import tethys_services.models
+
+
+class Migration(migrations.Migration):
```
swainn: Generate new migration after moving model.

tethys_config/models.py
```diff
@@ -41,3 +42,21 @@ def as_dict(cls):
...
+
+class AppProxyInstance(models.Model):
+    """
+    ORM for App Proxy which allows you to redirect an app to another host
+    """
+
+    name = models.CharField(max_length=100, unique=True)
+    endpoint = models.CharField(max_length=1024, validators=[validate_url])
+    logo_url = models.CharField(max_length=100, validators=[validate_url], blank=True)
+    description = models.TextField(max_length=2048, blank=True)
```
swainn: Add a field for tags like TethysApp.

tethys_apps/views.py
```diff
@@ -38,6 +39,22 @@ def library(request):
...
+    for app in proxied_apps:
+        new_app = {
+            'show_in_apps_library': True,
+            'enabled': True,
+            'proxied': True,
+            'url': getattr(app, 'endpoint'),
+            'icon': getattr(app, 'logo_url'),
+            'name': getattr(app, 'name'),
+            'description': getattr(app, 'description'),
+            'tags': []
```
swainn: Get tags from model.

swainn:
Merged into PR #463

rfun:
@swainn : Sorry for the delay on this. Seems like you have resolved all of the above changes before merging it in. Let me know if there is still something else left to do on this part.

Issue #434, opened 2019-06-18
JQuery v1.11.1 vulnerabilities


ezrajrice:
The current version of JQuery that is packaged with Tethys Platform contains two known vulnerabilities. One cross-site scripting (XSS) vulnerability and one prototype pollution vulnerability. Updating to version 3.4.0 or 3.4.1 (latest) will fix these issues.
See [link](https://snyk.io/test/npm/jquery/1.11.1) for more details on the version 1.11.1 vulnerabilities.

swainn:
Addressed by #441

Issue #433, opened 2019-06-18
Bootstrap v3.2.0 vulnerabilities


ezrajrice:
The current version of Bootstrap that is packaged with Tethys Platform contains five known cross-site scripting (XSS) vulnerabilities. Updating to version 3.4.1 or 4.3.1(latest) would fix these issues.
See [link](https://snyk.io/test/npm/bootstrap/3.2.0) for more details on the vulnerabilities.

swainn:
Upgrading to 4: https://getbootstrap.com/docs/4.0/migration/

swainn:
Addressed by #441

PR #441, opened 2019-07-09
Upgrade JQuery and Bootstrap

swainn:
Upgrades JQuery and Bootstrap to address known vulnerabilities as pointed out in #433 and #434. This is long overdue. We should consider cherry picking these into the release branch (2.1).


swainn:
@sdc50 @SarvaPulla this should be good to go assuming it passes the CI.

swainn:
> @sdc50 @SarvaPulla this should be good to go assuming it passes the CI.

Actually, disregard that. I want to rebase it with Scott's changes after we merge those in first.

docs/tethys_sdk/templating.rst
```diff
@@ -644,4 +500,12 @@ See the `Blocks`_ section for an explanation of each block.
...
-    {% endblock %}
\ No newline at end of file
+    {% endblock %}
+
+app_base.html
+=============
+
+This section provides the complete contents of the :file:`app_base.html` template. It is meant to be used as a reference for app developers, so they can be aware of the HTML structure underlying their app templates.
+
+.. literalinclude:: ../../tethys_apps/templates/tethys_apps/app_base.html
```
sdc50: Hurray for DRY!

tethys_apps/context_processors.py
```diff
@@ -41,4 +41,20 @@ def tethys_apps_context(request):
             if hasattr(app, 'enable_feedback'):
                 context['tethys_app']['enable_feedback'] = app.enable_feedback
 
+    # Dependency Versions
```
sdc50: I don't understand context processors as well as I should, but this is basically a pre-processor for all contexts before a template gets rendered, right? That's a good place to put this stuff. 👍

docs/tethys_sdk/templating.rst
swainn: Right?

tethys_apps/context_processors.py
swainn: Yes, that's the basic idea. I toyed with the idea of having the version in the settings.py... but I'm not sure it will always be as easy as switching the versions so I put it in there for now.

PR #443, opened 2019-07-11
conda package

sdc50:
Several changes needed to conda package tethys:

- Moved `static` and `template` directories and `manage.py` into `tethys_portal`.
- Rename `tests` to `tethys_tests` (needed because of moving `manage.py`)
- Add a CLI command to print the location of the `settings.py` file -- `tethys list -s` (I'm not sure this is the right place).
- add a conda recipe

(Also, fixes a database migration for `feedback_emails`)

I've uploaded a dev package to test. Here are some instructions for setting that up:

Conda Installation and Setup
====================

```
conda create -n tethys-pkg -c tethysplatform -c tethysplatform/label/dev -c conda-forge --use-local tethysplatform
conda activate tethys-pkg
tethys gen settings
tethys list -s
# edit any settings that you want
tethys db configure
tethys manage start
```


swainn:
I got an error after running the `tethys db configure` (full output below). It took me a while to realize that I had another Tethys database running on the same port, but I think it would be nearly impossible for normal users to deduce that based on the output that is given. I think we need to catch these errors and give better feedback to the users. There were two things that clued me into what was really going on:

1. The database couldn't start: `pg_ctl: could not start server`
2. The error indicates objects in the database already existing

I recommend we try to catch errors with starting the db when running the `tethys db configure` command and stop execution if it is unable to start the db, and display a user friendly message like "Unable to start database, is there a database already running on port XXXX?"

It would also be helpful to log something to the console  in blue or a color that sticks out to indicate what is going on under the covers (e.g.: "Initializing database...", "Starting database server...", "Running migrations...", etc.). That way it would be more easy to connect errors to operations.

That said, once I stopped the offending database, it worked like a charm.

Full error output:
```
tethys db configure
The files belonging to this database system will be owned by user "nswain".
This user must also own the server process.

The database cluster will be initialized with locale "en_US.UTF-8".
The default database encoding has accordingly been set to "UTF8".
The default text search configuration will be set to "english".

Data page checksums are disabled.

creating directory /home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/conda-pkg-psql/data ... ok
creating subdirectories ... ok
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default timezone ... UTC
selecting dynamic shared memory implementation ... posix
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
syncing data to disk ... ok

WARNING: enabling "trust" authentication for local connections
You can change this by editing pg_hba.conf or using the option -A, or
--auth-local and --auth-host, the next time you run initdb.

Success. You can now start the database server using:

    pg_ctl -D /home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/conda-pkg-psql/data -l logfile start

waiting for server to start.... stopped waiting
pg_ctl: could not start server
Examine the log output.
ERROR:  role "tethys_default" already exists
createdb: database creation failed: ERROR:  database "tethys_platform" already exists
ERROR:  role "tethys_super" already exists
/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/distributed/utils.py:139: RuntimeWarning: Couldn't detect a suitable IP address for reaching '8.8.8.8', defaulting to '127.0.0.1': [Errno 101] Network is unreachable
  RuntimeWarning,
Loading Tethys Extensions...
Loading Tethys Apps...
Operations to perform:
  Apply all migrations: admin, auth, authtoken, captcha, contenttypes, guardian, sessions, social_django, termsandconditions, tethys_apps, tethys_compute, tethys_config, tethys_quotas, tethys_services
Running migrations:
  No migrations to apply.
/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/distributed/utils.py:139: RuntimeWarning: Couldn't detect a suitable IP address for reaching '8.8.8.8', defaulting to '127.0.0.1': [Errno 101] Network is unreachable
  RuntimeWarning,
Loading Tethys Extensions...
Loading Tethys Apps...
Traceback (most recent call last):
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "auth_user_username_key"
DETAIL:  Key (username)=(admin) already exists.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/nswain/miniconda3/envs/tethys-pkg/bin/tethys", line 11, in <module>
    load_entry_point('tethys-platform==3.0.0a1', 'console_scripts', 'tethys')()
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/tethys_cli/__init__.py", line 55, in tethys_command
    args.func(args)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/tethys_cli/db_commands.py", line 153, in db_command
    DB_COMMANDS[args.command](**options)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/tethys_cli/db_commands.py", line 114, in configure_tethys_db
    create_portal_superuser(**kwargs)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/tethys_cli/db_commands.py", line 105, in create_portal_superuser
    User.objects.create_superuser(portal_superuser_name, portal_superuser_email, portal_superuser_password)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/contrib/auth/models.py", line 161, in create_superuser
    return self._create_user(username, email, password, **extra_fields)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/contrib/auth/models.py", line 144, in _create_user
    user.save(using=self._db)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/contrib/auth/base_user.py", line 73, in save
    super().save(*args, **kwargs)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/models/base.py", line 718, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/models/base.py", line 748, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/models/base.py", line 831, in _save_table
    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/models/base.py", line 869, in _do_insert
    using=using, raw=raw)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/models/query.py", line 1136, in _insert
    return query.get_compiler(using=using).execute_sql(return_id)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1289, in execute_sql
    cursor.execute(sql, params)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/backends/utils.py", line 100, in execute
    return super().execute(sql, params)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/backends/utils.py", line 68, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/backends/utils.py", line 77, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/utils.py", line 89, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/home/nswain/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.IntegrityError: duplicate key value violates unique constraint "auth_user_username_key"
DETAIL:  Key (username)=(admin) already exists.
```

swainn:
The default location for the database directory to be installed for the conda-package is the pythonX.X directory: `~/miniconda3/envs/tethys-pkg/lib/python3.7/psql`. Is that where we want it installed by default? Seems something like `~/tethys/psql` would be better. Or even in `~/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/tethysplatform`. What is the default location for the postgresql conda package?

swainn:
The `templates` and `static` directories are missing for all `tethys` packages except `tethys_portal`.

List of all items with `templates`  in name in directories starting with `tethys`:
```
ls -l .../python3.7/site-packages/tethys_*/ | grep "/\|templates"
.../python3.7/site-packages/tethys_apps/:
.../python3.7/site-packages/tethys_cli/:
drwxr-xr-x 3 user user  4096 Jul 11 17:50 gen_templates
drwxr-xr-x 4 user user  4096 Jul 11 17:50 scaffold_templates
.../python3.7/site-packages/tethys_compute/:
.../python3.7/site-packages/tethys_config/:
.../python3.7/site-packages/tethys_dask_scheduler/:
.../python3.7/site-packages/tethys_dask_scheduler-1.0.2.dist-info/:
.../python3.7/site-packages/tethys_dataset_services/:
.../python3.7/site-packages/tethys_dataset_services-2.0.0-py3.7.egg-info/:
.../python3.7/site-packages/tethys_gizmos/:
.../python3.7/site-packages/tethys_platform-3.0.0a1-py3.6.egg-info/:
.../python3.7/site-packages/tethys_portal/:
drwxr-xr-x 5 user user  4096 Jul 11 17:50 templates
.../python3.7/site-packages/tethys_quotas/:
.../python3.7/site-packages/tethys_sdk/:
.../python3.7/site-packages/tethys_services/:
```

List of all items with `static`  in name in directories starting with `tethys`:
```
user user  2558 Jul  8 11:17 static_finders.py
.../python3.7/site-packages/tethys_cli/:
.../python3.7/site-packages/tethys_compute/:
.../python3.7/site-packages/tethys_config/:
.../python3.7/site-packages/tethys_dask_scheduler/:
.../python3.7/site-packages/tethys_dask_scheduler-1.0.2.dist-info/:
.../python3.7/site-packages/tethys_dataset_services/:
.../python3.7/site-packages/tethys_dataset_services-2.0.0-py3.7.egg-info/:
.../python3.7/site-packages/tethys_gizmos/:
.../python3.7/site-packages/tethys_platform-3.0.0a1-py3.6.egg-info/:
.../python3.7/site-packages/tethys_portal/:
drwxr-xr-x 3 user user  4096 Jul 11 17:50 static
.../python3.7/site-packages/tethys_quotas/:
.../python3.7/site-packages/tethys_sdk/:
.../python3.7/site-packages/tethys_services/:
```

sdc50:
> The default location for the database directory to be installed for the conda-package is the pythonX.X directory: `~/miniconda3/envs/tethys-pkg/lib/python3.7/psql`. Is that where we want it installed by default? Seems something like `~/tethys/psql` would be better. Or even in `~/miniconda3/envs/tethys-pkg/lib/python3.7/site-packages/tethysplatform`. What is the default location for the postgresql conda package?

I agree that the default location is not ideal. If you specify an absolute path for the `DIR` setting in `settings.py` then it will be installed there. If the path is relative then it is relative to `TETHYS_HOME`. Currently it defaults to a relative directory called `psql`. With the conda install `TETHYS_HOME` is not set, so the default for that is the parent directory of `TETHYS_SRC`.  I think we should change the default location of `TETHYS_HOME` to be `~/tethys` or `~/.tethys`. If a user wants it in a different location they can specify the `TETHYS_HOME` environment variable.

sdc50:
> I think we should change the default location of `TETHYS_HOME` to be `~/tethys` or `~/.tethys`.

I'm leaning toward `~/.tethys`. Any other thoughts?

sdc50:
> I recommend we try to catch errors with starting the db when running the `tethys db configure` command and stop execution if it is unable to start the db, and display a user friendly message like "Unable to start database, is there a database already running on port XXXX?"

I'm not sure that I can create a better error message than what is output by postgres. I can catch the error and exit so that the error messages are more obvious. Anything that I add to the error message would just be speculation.

swainn:
> I'm leaning toward `~/.tethys`. Any other thoughts?

I like ~/.tethys

sdc50:
I didn't like the `tethys list -s` command for printing the path to the settings.py file at first, but since I can't think of anything better, it's starting to grow on me. I think this PR is almost ready to go. I just have a few remaining questions:

1. Should the conda package have an activate script that sets any environment variables or aliases?
2. Should we adjust the CI to test building the conda package?  and, if so, should the tests run against the built package or against the source code?

sdc50:
BTW, I uploaded another package with the latest state of this PR to tethysplatform/label/dev.

swainn:
> 1. Should the conda package have an activate script that sets any environment variables or aliases?

I think we ought to set all the aliases that people are used to using (tms, tstartdb, etc.). Seems that TETHYS_SRC/TETHYS_HOME needed to be set too, no?

> 2. Should we adjust the CI to test building the conda package?  and, if so, should the tests run against the built package or against the source code?

Yes for testing the conda build. I think we ought to run the tests for both the source and the conda build. To ensure consistency. What do you think?

conda.recipe/build.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+python setup.py install --single-version-externally-managed --record=record.txt
+
+# Copy non-.py files
+
+# tethys_apps
+cp -r $SRC_DIR/tethys_apps/static $SP_DIR/tethys_apps/
```
swainn: Is there a reason you are manually copying the resource files here? If not, I think it would be better to have the setup.py do it. We have a utility function that can be used to find the resource files. Take a look at how the app setup.py do it: https://github.com/tethysplatform/tethys/blob/master/tethys_cli/scaffold_templates/app_templates/default/setup.py_tmpl

conda.recipe/build.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# tethys_quotas
+cp -r $SRC_DIR/tethys_quotas/templates $SP_DIR/tethys_quotas/
+
+# tethys_services
+cp -r $SRC_DIR/tethys_services/static $SP_DIR/tethys_services/
+cp -r $SRC_DIR/tethys_services/templates $SP_DIR/tethys_services/
+
+
+rm -rf $SP_DIR/tethys_tests/
```
swainn: This could also be handled in the setup.py by just not listing/ignoring the `tethys_tests` package.

conda.recipe/build.sh
```diff
@@ -0,0 +1,36 @@
...
+rm $SP_DIR/tethys_portal/settings.py*
```
swainn: I'm guessing this is a safe guard for local builds?

conda.recipe/meta.yaml
```diff
@@ -0,0 +1,105 @@
...
+  version: {{ data['version'] }}
+
+
+source:
+  path: ..
+
+build:
+  number: 0
+  noarch: python
+#  script:  python setup.py install --single-version-externally-managed --record=record.txt
```
swainn: Remove commented code.

conda.recipe/meta.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
+  run:
+    - python >=3.6
+
+    # system dependencies
+    - pycrypto
```
swainn: We discussed locking the versions of dependencies for the conda build. What's the most straight forward way to do that?

conda.recipe/build.sh
swainn: I think we ought to handle the resources through the setup.py so that it installs manually the same way it does in the conda environment. The app setup.py use a utility function for finding all the resource files. See: https://github.com/tethysplatform/tethys/blob/master/tethys_cli/scaffold_templates/app_templates/default/setup.py_tmpl

conda.recipe/build.sh
swainn: The tethys_tests directory ought to be excluded in the setup.py as well.

conda.recipe/build.sh
swainn: I assume this is a safe guads for local builds?

conda.recipe/meta.yaml
swainn: Have we made a decision on whether these should be locked down or not? Perhaps this file should be generated with options to lock all dependencies for release builds, or leave it open for dev builds?

sdc50:
> I think we ought to set all the aliases that people are used to using (tms, tstartdb, etc.). Seems that TETHYS_SRC/TETHYS_HOME needed to be set too, no?

I think we should drop all of the alias for the database since we now have the `tethys db` CLI. I also don't think we need to have TETHYS_SRC/TETHYS_HOME set here; the defaults for those are fine. 

The `tms` command and possibly the `tstart` commands are the only ones that I think make sense to have.

swainn:
> > I think we ought to set all the aliases that people are used to using (tms, tstartdb, etc.). Seems that TETHYS_SRC/TETHYS_HOME needed to be set too, no?
> 
> I think we should drop all of the alias for the database since we now have the `tethys db` CLI. I also don't think we need to have TETHYS_SRC/TETHYS_HOME set here; the defaults for those are fine.
> 
> The `tms` command and possibly the `tstart` commands are the only ones that I think make sense to have.

With that reasoning, are any of the aliases necessary? We have the `tethys manage start` CLI.

conda.recipe/build.sh
sdc50: I've moved all of the resource file handling into a `MANIFEST.in` file. The `include_package_data=True` option in the `setup.py` will look for the `MANIFEST.in` to determine which package data to include.

conda.recipe/build.sh
sdc50: I've added code in the `setup.py` that temporarily renames the `settings.py` file to prevent it from being included in a local build/install. I couldn't work out any other way of having a single `.py` file excluded.

conda.recipe/meta.yaml
sdc50: We just need to add all of the version pins into the `meta.yaml` file. Alternatively we could add all of these dependencies to the `setup.py` file and then copy them from there for the conda recipe.

sdc50: This is an interesting idea. Depending on if we want to release a PyPI package then we'll need to maintain several versions of the dependencies. Having them in a template might be a good way to consolidate that.

sdc50:
> With that reasoning, are any of the aliases necessary? We have the `tethys manage start` CLI.

No, at this point I don't think any of the aliases are necessary. It's just a matter of if they are enough of a convenience for our users to include them as standards. I personally think we should just drop them all together. Users can set up aliases themselves if they want them.

sdc50:
> Yes for testing the conda build. I think we ought to run the tests for both the source and the conda build. To ensure consistency. What do you think?

I think in order to run tests against the conda build we'll have to include the tests as one of the installed packages, which I don't like. However, testing is part of our CLI, so that would break if we don't include the tests package. Maybe tests shouldn't be part of the CLI at all.

swainn:
@sdc50 I tried installing the conda package on windows. The environment seemed to install just fine, but the tethys command is not recognized after activating the environment. I've confirmed that the `tethysplatform` package is installed and there is a script in the environments Scripts directory called tethys, but Windows isn't recognizing it as a command. Even if I'm in the Script directory.

If I run `python tethys <some command>` I get the following Traceback:

```
D:\miniconda3\envs\tethys-pkg\Scripts
>python tethys gen settings
Traceback (most recent call last):
  File "tethys", line 11, in <module>
    load_entry_point('tethys-platform==3.0.0a2', 'console_scripts', 'tethys')()
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\pkg_resources\__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\pkg_resources\__init__.py", line 2843, in load_entry_point
    return ep.load()
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\pkg_resources\__init__.py", line 2434, in load
    return self.resolve()
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\pkg_resources\__init__.py", line 2440, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\tethys_cli\__init__.py", line 13, in <module>
    from tethys_cli.app_settings_commands import add_app_settings_parser
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\tethys_cli\app_settings_commands.py", line 1, in <module>
    from tethys_apps.utilities import get_app_settings
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\tethys_apps\utilities.py", line 15, in <module>
    from tethys_apps.harvester import SingletonHarvester
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\tethys_apps\harvester.py", line 16, in <module>
    from tethys_apps.base import TethysAppBase, TethysExtensionBase
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\tethys_apps\base\__init__.py", line 11, in <module>
    from tethys_apps.base.app_base import TethysAppBase, TethysExtensionBase  # noqa: F401
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\tethys_apps\base\app_base.py", line 21, in <module>
    from tethys_apps.base.testing.environment import is_testing_environment, get_test_db_name, TESTING_DB_FLAG
  File "D:\miniconda3\envs\tethys-pkg\lib\site-packages\tethys_apps\base\testing\environment.py", line 1, in <module>
    from os import environ, unsetenv
ImportError: cannot import name 'unsetenv' from 'os' (D:\miniconda3\envs\tethys-pkg\lib\os.py)
```

sdc50:
> ImportError: cannot import name 'unsetenv' from 'os' (D:\miniconda3\envs\tethys-pkg\lib\os.py)

That is strange, because the docs say that `unsetenv` is supported on Windows. However, it also says that it is deleting `os.environ` keys is preferred to calling `unsetenv`, which the code is already doing, so I just removed that line and import statement.

sdc50:
@swainn This should be ready to go. I've updated the install docs. I think there is some work work that could be done to update the docs, but I think that should be done as a separate PR.

PR #455, opened 2019-07-22
Overview map

sdc50:
Add overview map control to `tethys_map_view`

### DIFF
--- a/tethys_gizmos/templatetags/tethys_gizmos.py
+++ b/tethys_gizmos/templatetags/tethys_gizmos.py
@@ -77,6 +77,39 @@ class HighchartsDateEncoder(DjangoJSONEncoder):
         return super().default(obj)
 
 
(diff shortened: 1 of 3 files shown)


tethys_gizmos/templatetags/tethys_gizmos.py
```diff
@@ -77,6 +77,39 @@ def default(self, obj):
...
+
+        for name in self.var_names:
+            context = context[name]
+
+        context[self.var_name] = value
+
+        return ''
+
+
+@register.tag(name='set')
```
swainn: Is there a reason you don't want to use the `with` tag to set variables in the template?

swainn:
> Will you take care of cherry-picking this into dev?

* Are you going to patch this into release as well?

swainn:
NVM. I see there are 2 PR's now. One for master and one for release.

PR #463, opened 2019-08-08
Django 2.2, Proxy Apps, and Various Bug Fixes

swainn:
* Updates Tethys to support Django 2.2
* Refactors `tethys install` to support installing extensions
* `tethys install` will attempt to install app and extension dependencies without update already installed dependencies by default. This can be overridden with the ``--update-installed`` option.
* Adds `--only-dependencies` option to install command to allow for installing only the dependencies.
* Adds `--without-dependencies` option to install command to allow for installing without the dependencies from the CLI.
* Fixes many issues with the initial implementation of `tethys install`
* Fixes broken Dask Dashboard link in admin pages
* Constrains the Dask version to the version that works with Tethys - non-trivial upgrade...
* Fixes issue with draw ids not being unique on MapView.
* Improve the behavoir of the Jobs Table to be more user-friendly and dynamic.
* Fixes bug with workflow node graph not working with remote condor schedulers.
* Adds ProxyApp support.


swainn:
@sdc50 @SarvaPulla this PR includes the ProxyApp contribution from @rfun.

tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
```diff
@@ -1477,9 +1472,28 @@ var TETHYS_MAP_VIEW = (function() {
    * Attribute Table Methods
    ***********************************/
   generate_feature_id = function() {
-    var id = m_draw_id_counter;
-    ++m_draw_id_counter;
-    return id;
+    //// return uuid of form xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx
```
sdc50: What is this for?

PR #467, opened 2019-08-27
Automated CLI Documentation

swainn:
* Automatically generate CLI documentation


swainn:
@SarvaPulla @sdc50 this is ready for merge. I've refactored the docs to automatically generate the CLI documentation the argparse help. A few notes on this:

1. I've moved the CLI documentation out of the SDK documentation to the root level so it is more easily accessible. Let me know what you think of that.
2. The approach I took was to create one page for each subcommand. The tool can parse subcommands recursively and generate it all in one page, but after a few nested levels of headers, it becomes hard to read/navigate the documentation. Splitting them out to separate pages for each command makes it a lot easier to read at the cost of a little more work on our end to maintain it. 
3. You'll probably notice that I've hardcoded descriptions of each subcommand b/c for some reason, the library doesn't generate the description when you generate docs for a single subcommand. Perhaps a suggestion for them in the future...

swainn:
@SarvaPulla @sdc50 one other thing on this: I've been seriously considering changing the theme of our documentation, partially just for a change, but also, I think it would be helpful to make it obvious when you are looking a the Tethys 3 docs vs. older versions of the docs (I think we should maintain the Tethys 2 docs for some time after we release 3 for those that are still using the old version). What do you think?

If you agree that we should change it, I'd appreciate some recommendations from you on which theme. Take a look at this: https://sphinx-themes.org/

SarvaPulla:
I like sphinx_drove_theme or sphinx-ustack-theme. We could try to update the look for 3.0.

Issue #480, opened 2019-09-13
Fix django-guardian version on stable environment_py files


rfun:
Installing from the latest release will give an error as django-guardian's latest version (released sep 8th) required path from django.urls, which doesn't exist before django 2. Fixing the version to django-guardian=1.5.* does the trick. We probably want to update the stable environment files accordingly.

rileyhales:
I also ran into this problem yesterday and ended up doing the same thing to fix it.

swainn:
Just released a new version of Tethys: https://github.com/tethysplatform/tethys/releases/tag/2.1.6

Please verify.

rileyhales:
I just tried a fresh install on an ubuntu virtual machine and it solved the problem. i could start the development server and install an app. looks good to me.

PR #482, opened 2019-09-16
Fix django-guardian version

rileyhales:
the fix suggested in response to Issue #480 and tested by myself and @rfun

### DIFF
--- a/stable_environment_py3.yml
+++ b/stable_environment_py3.yml
@@ -43,7 +43,7 @@ dependencies:
 - djangorestframework
 - django-bootstrap3
 - django-model-utils
-- django-guardian
+- django-guardian=1.5
 
 # tests dependencies
 - requests-mock


rileyhales:
Travis failed but if you check the details it had the same errors on the python 2 and 3 versions but only failed on 3. it also worked for me locally and apparently also for rohit. not sure whats up with that.

swainn:
@rileyhales @rfun can you confirm if this issue is in the master branch? Do we need to merge this into that branch as well?

rileyhales:
@swainn i just tried a fresh install of tethys 3.0.0 from the tethysplatform/label/dev channel using http://docs.tethysplatform.org/en/latest/installation.html instructions and it seems to be fine. i installed, generated settings, configured the db, and started a local server without any problems.

swainn:
@sdc50 and @SarvaPulla the tests that are failing b/c of a change with the latest version of condorpy. I'm looking into that issue now, but it seems to be unrelated to these changes. I recommend we merge these changes and release a patch version, but I'd like at least one of you to take a look and double-check my reasoning.

swainn:
> @swainn i just tried a fresh install of tethys 3.0.0 from the tethysplatform/label/dev channel using http://docs.tethysplatform.org/en/latest/installation.html instructions and it seems to be fine. i installed, generated settings, configured the db, and started a local server without any problems.

Thanks @rileyhales

swainn:
Sorry this took so long.

PR #495, opened 2019-10-05
Settings yaml

sdc50:


tethys_portal/settings.py
```diff
@@ -12,41 +12,76 @@
```
swainn: I think we ought to add a big warning at the top of this file, reminding people not to edit this file directly and pointing them to the documentation for configuring settings.

docs/tethys_portal/configuration.rst
```diff
@@ -0,0 +1,372 @@
...
+----------------
+
+The following is a list of keys that can be added to the :file:`portal.yml` file:
+
+* ``version``: The version of the :file:`portal.yml` file schema
+* ``name``:
+* ``apps``:
+* ``settings``:
+
+  * ``SECRET_KEY``:
```
swainn: Are you planning to add descriptions for these? I'd be happy to help out here if you just want to get it merged.

Dockerfile
```diff
@@ -29,7 +29,8 @@ ENV  BASH_PROFILE=".bashrc" \
      CLIENT_MAX_BODY_SIZE="75M"
 
 # Tethys settings arguments
-ENV  ALLOWED_HOSTS="localhost 127.0.0.1" \
+ENV  DEBUG="True" \
```
swainn: I see the Dockerfile being primarily used for production, so I would recommend having the default values be appropriate for a production deployment if possible (i.e.: DEBUG="False").

docs/installation.rst
```diff
@@ -32,24 +32,15 @@ To install the ``tethysplatform`` into a new conda environment then run the foll
...
 ------------------------------------
 
-Tethys requires a :file:`settings.py` file to configure the database and other local settings. To generate a new :file:`settings.py` with default values for your local installation run::
+To add custom configurations such as the database and other local settings you will need to generate a :file:`portal.yml` file. To generate a new template :file:`portal.yml` run::
 
-    tethys gen settings
+    tethys gen portal
 
-You can customize your :file:`settings.py` file when you generate it by passing in command line options. Refer to the :ref:`tethys_gen_cmd` docs to see what options are available.
+You can customize your settings in the :file:`portal.yml` file after you generate it by manually editing the file or by using the :ref:`tethys_settings_cmd` command. Refer to the :ref:`tethys_configuration` documentation for me information.
```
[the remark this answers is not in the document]
sdc50: Good catch

swainn:
I just thought of another comment: Can you add an entry to the What's New page for the new settings approach?

swainn:
Also, what do you think about having `tethys settings --get` accept no argument as an option to return all of the settings? And can we pretty format the output of that command?

docs/tethys_portal/configuration.rst
sdc50: Yes, I could definitely use some help! I meant to add descriptions to each key, but I'm now way overdue to be done with this and have some other deadlines that I need to scramble to hit.

tethys_cli/settings_commands.py
```diff
@@ -0,0 +1,128 @@
...
+        nargs=1,
+    )
+    settings_parser.set_defaults(
+        func=settings_command,
+    )
+
+
+def read_settings():
+    portal_yaml_file = TETHYS_HOME / 'portal.yml'
+    tethys_settings = yaml.safe_load(portal_yaml_file.open()).get('settings', {}) or {}
```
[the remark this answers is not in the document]
sdc50: I'm not positive that it does. I'll edit to the code to be safe.

tests/apps/tethysapp-test_app/test-site-content.yml
```diff
@@ -1,32 +1,32 @@
 # site_content.yml
```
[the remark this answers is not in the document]
sdc50: Yes, but it should be called test-portal.yml now to be more accurate.

Dockerfile
sdc50: Yes, that was my intention. Oops!

sdc50:
> Do changes to the portal.yml, either through editing or using the commands, cause an automatic restart of the Django server?

No, I don't think it will.

sdc50:
> I just thought of another comment: Can you add an entry to the What's New page for the new settings approach?

I was planning to do this, but it looks like someone already has.

sdc50:
> Also, what do you think about having `tethys settings --get` accept no argument as an option to return all of the settings? And can we pretty format the output of that command?

Done!

docs/tethys_portal/configuration.rst
swainn: Let’s go ahead and merge for now.

PR #506, opened 2019-12-13
THREDDS Integration

swainn:
* This is derived from @rfun 's PR #439
* THREDDS is added as a SpatialDatasetService, instead of a separate THREDDS service
* Adds Unidata's image to the Docker CLI


swainn:
@SarvaPulla, @sdc50 this should be ready to go.

tethys_apps/models.py
```diff
@@ -295,7 +295,8 @@ def get_value(self):
                 self.value = self.default
 
         if self.value == '' or self.value is None:
-            return None  # TODO Why don't we raise a NotAssigned error here?
+            # None is a valid value to return in the case the value has not been set for this setting type
+            return None
```
sdc50: It isn't clear to me why a `TethysAppSettingNotAssigned` error isn't raised here.

swainn: This is the ``get_value`` method for the custom settings, which are not really assigned to. They take simple values like a boolean or integer, whereas the other settings are "assigned" service objects. So in my mind the NotAssigned error only applies to services that are assigned service objects and empty is a valid state for simple valued settings.

swainn:
@rfun we've finally merged your THREDDS integration PR with some changes.

PR #514, opened 2020-01-16
Fix for #504

### LINKED ISSUE #504: Inconsistent Problems Arise When Installing New Version of Tethys

sdc50:
workaround fix for the intermittent Django startup errors. Fixes #504

### DIFF
--- a/scripts/install_tethys.sh
+++ b/scripts/install_tethys.sh
@@ -361,7 +361,7 @@ then
     then
         tethys gen portal_config
         tethys settings \
-          --set ALLOWED_HOSTS "${ALLOWED_HOST}" \
+          --set ALLOWED_HOSTS "[${ALLOWED_HOST}]" \
           --set DATABASES.default.USER ${TETHYS_DB_USERNAME} \
           --set DATABASES.default.PASSWORD ${TETHYS_DB_PASSWORD} \
           --set DATABASES.default.PORT ${TETHYS_DB_PORT} \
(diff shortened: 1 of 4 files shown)


scripts/install_tethys.sh
```diff
@@ -530,9 +522,9 @@ then
 
     echo "export NGINX_USER='${NGINX_USER}'" >> "${ACTIVATE_SCRIPT}"
     echo "export NGINX_HOME='${NGINX_HOME}'" >> "${ACTIVATE_SCRIPT}"
-    echo "alias tethys_user_own='sudo chown -R \${USER} \"\${TETHYS_SRC}\" \"\${TETHYS_HOME}/static\" \"\${TETHYS_HOME}/workspaces\" \"\${TETHYS_HOME}/apps\"'" >> "${ACTIVATE_SCRIPT}"
+    echo "alias tethys_user_own='sudo chown -R \${USER} \"${TETHYS_SRC}\" \"${TETHYS_HOME}/static\" \"${TETHYS_HOME}/workspaces\" \"${TETHYS_HOME}/apps\"'" >> "${ACTIVATE_SCRIPT}"
```
swainn: Are ``TETHYS_SRC`` and ``TETHYS_HOME`` defined in this script still?

swainn: The SERVIR folks rely pretty heavily on these aliases for managing their manual deploys.

sdc50: Yes, both `TETHYS_SRC` and `TETHYS_HOME` are defined in the install script, but I have removed them from the activate script. The changes here have the effect of substituting the values for those variables when the activate scripts are written, rather than writing the variable names in the script so they are substituted at runtime.

sdc50:
> The ALLOWED_HOSTS section of the portal_config.yml isn't formatted correctly when installing from the script.

It looks like the script was not passing in the `ALLOWED_HOSTS` argument as a list. This should be fixed now.

PR #521, opened 2020-01-27
Enhance db command

sdc50:
add purge db command
enhance configure db command so it prompts on error

### DIFF
--- a/tethys_cli/db_commands.py
+++ b/tethys_cli/db_commands.py
@@ -8,8 +8,10 @@
 ********************************************************************************
 """
 from pathlib import Path
+import shutil
 
 from django.conf import settings
+from django.db.utils import IntegrityError
 
 from tethys_cli.cli_helpers import get_manage_path, run_process, load_apps
 from tethys_cli.cli_colors import write_info, write_error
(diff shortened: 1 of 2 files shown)


tethys_cli/db_commands.py
```diff
@@ -22,7 +24,20 @@ def add_db_parser(subparsers):
...
+                                '  * start - Starts the local database server.\n'
+                                '  * stop - Stops the local database server.\n'
+                                '  * status - Gets the current status of the local database server.\n'
+                                '  * create - Creates the Tethys Portal database on the database server configured '
+                                'in the `portal-config.yml` (this could be local or remote).\n'
+                                '  * migrate - Runs migrations on the Tethys Portal database.\n'
+                                '  * createsuperuser - Creates a Tethys Portal superuser account.\n'
+                                '  * configure - A shortcut for running: init, start, create, migrate, and '
+                                'createsuperuser. Note: if the database server that is configured in '
+                                '`portal_config.yml` is remote then the init and start commands are skipped.\n'
```
swainn: Technically a docker container or system database could be running locally, but you wouldn't want to run the init or start  steps. For clarity, this should probably say something to the effect of "if the DIR parameter is not defined in your DATABASES.default configuration in portal_config.yml then the init and start commands are skipped."

PR #522, opened 2020-02-04
Fixes for cesium map views

gagelarsen:
- Added cesium token parameter
- Updated the URL for cesium dependencies
- Changed layers and entities to take lists
- Updated the default version of cesium to 1.63.1
- Added support for GeoJSON entities
- Added support for TileWMS and ImageWMS to MVLayers for layers
- Added support for GeoJSON to MVLayers for entities
- Added a GeoJSON example for cesium to the gizmo showcase
- Added input of cesium token to the gizmo showcase

### DIFF
--- a/docs/tethys_sdk/gizmos/map_view.rst
+++ b/docs/tethys_sdk/gizmos/map_view.rst
@@ -6,6 +6,8 @@ Map View
 
 .. autoclass:: tethys_sdk.gizmos.MapView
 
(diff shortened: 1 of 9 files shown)


swainn:
@sdc50 and @SarvaPulla I need you to review this PR since I was involved in its development. These are some changes we needed to make to Cesium to get it working in our workflows. The current version in Tethys is no longer supported due to new licensing requirements.

PR #529, opened 2020-02-26
add example code for opening ssl port

rileyhales:
it took me a while to figure this one out. i forgot the production install had me use this firewall utility. I think it would be wise to have the SSL instructions mirror the standard http instructions so save confusion like this.


swainn:
@rileyhales this looks good to me. D

PR #533, opened 2020-03-11
Fix bug with scaffolding extensions.

swainn:
* Scaffolding extensions was not working properly
* It would throw an error
* Looks like the templates for extensions were not updated when the templating engine was changed from Django to Jinja2 again.


swainn:
This fix is included in PR #536 which also fixes the issues that are causing the tests to fail.

PR #536, opened 2020-03-16
Bokeh 2.0.0 updates and Scaffold extension fix

msouff:


sdc50:
l've been playing with the Bokeh Tutorial app to test the Bokeh integration with Tethys. I have an environment with:
* Tethys - master branch
* Bokeh = 1.4
* Panel = 0.8.0
The sea surface example worked, but the shapes did not. I kept getting the following error (in the browser console):

```
Error rendering Bokeh items: Error: Model 'panel.models.markup.HTML' does not exist. This could be due to a widget or a custom model not being registered before first usage.
```

I tried updating to Bokeh= 2.0, which caused both examples to not work. After applying the changes from this PR I was again able to the the sea surface example to work, but I'm continuing to get the same error with the shapes example. 

@msouff Do you have any ideas what might be wrong? Have you gotten it to work?

docs/tethys_sdk/url_maps.rst
```diff
@@ -107,7 +107,7 @@ The example below adds a column layout containing a slider and a plot. A callbac
                 data['y'] = [0, 10, 20, 30]
             else:
                 data['y'] = [i * new for i in [0, 10, 20, 30]]
-            source.data = ColumnDataSource(data=data).data
+            source.data = dict(ColumnDataSource(data=data).data)
```
swainn: @msouff is this the recommended approach in the new version of Panel? What is the reason for casting this to a dictionary?

environment.yml
```diff
@@ -9,7 +9,7 @@ channels:
 - tethysplatform
 - tethysplatform/label/dev
 - conda-forge
-- bokeh/label/dev
+- bokeh
```
swainn: @msouff Do we need the bokeh channel? I believe our goal is to eventually move to conda-forge exclusively.

msouff: 2.0.0 is available from conda-forge now. I think when I fixed this bokeh 2.0.0 was coming from the bokeh dev channel and I noticed it was also in the main bokeh channel and that's probably why I got rid of the bokeh dev channel. But now that it is in conda-forge too, we don't need that one either.

docs/tethys_sdk/url_maps.rst
msouff: Yes. This is the recommended now. The data parameter of ColumDataSource only takes plain Python dicts now. See the error below and the recommended change.

ValueError('\nColumnDataSource.data properties may only be set from plain Python dicts,\nnot other ColumnDataSource.data values.\n\nIf you need to copy set from one CDS to another, make a shallow copy by\ncalling dict: s1.data = dict(s2.data)\n'

msouff: Ahh this just made me realize why the bokeh tutorial is not working for Scott :-D

msouff:
> l've been playing with the Bokeh Tutorial app to test the Bokeh integration with Tethys. I have an environment with:
> 
> * Tethys - master branch
> * Bokeh = 1.4
> * Panel = 0.8.0
>   The sea surface example worked, but the shapes did not. I kept getting the following error (in the browser console):
> 
> ```
> Error rendering Bokeh items: Error: Model 'panel.models.markup.HTML' does not exist. This could be due to a widget or a custom model not being registered before first usage.
> ```
> 
> I tried updating to Bokeh= 2.0, which caused both examples to not work. After applying the changes from this PR I was again able to the the sea surface example to work, but I'm continuing to get the same error with the shapes example.
> 
> @msouff Do you have any ideas what might be wrong? Have you gotten it to work?

@sdc50 I just updated a minor syntax change from Bokeh 2.0.0 in the bokeh tutorial app. Try it with the latest stuff now (bokeh 2.0.0 + panel 0.9+). It works for me!

PR #542, opened 2020-03-27
Security Updates

ezrajrice:
This addresses security vulnerabilities: Password Field with Autocomplete Enabled, Cache-able HTTPS Response on content containing credentials, and Directory Enumeration in production settings.


ezrajrice:
@swainn @sdc50 Running $ tethys test -c -u appears to not use the configured portal_config.yml file. Any idea what it's using instead? Does it create a mock settings file or go with default parameters?

ezrajrice:
@swainn @sdc50 Continuing from previous comment, pulling in the settings and printing getattr(settings, 'DEBUG') shows that the tests are run with DEBUG = False. This does not match my portal_config.yml file. This explains the failed test_permission_required_exception_403 because with DEBUG = False, the method returns 404 instead of 403.

PR #543, opened 2020-03-27
Dev

ezrajrice:
This addresses security vulnerabilities: Password Field with Autocomplete Enabled, Cache-able HTTPS Response on content containing credentials, and Directory Enumeration in production settings.


ezrajrice:
Didn't include latest commits.

ezrajrice:
Nevermind, I'm just not used to Pull Requests. Sorry.

tethys_apps/decorators.py
```diff
@@ -102,7 +102,8 @@ def my_controller(request):
     """  # noqa: E501
 
     use_or = kwargs.pop('use_or', False)
-    message = kwargs.pop('message', "We're sorry, but you are not allowed to perform this operation.")
+    message = kwargs.get('message', "We're sorry, but you are not allowed to perform this operation.")
+    production_message = kwargs.get('message', "We're sorry, but the operation you requested cannot be found.")
```
sdc50: I think just having one message is sufficient. I don't think to distinguish between dev and production mode.

ezrajrice: I have made the suggested change.

Issue #540, opened 2020-03-23
Panel apps not working with Bokeh 2.0


sdc50:
The bokeh_tutorial app has both a Bokeh app and a Panel app example. I can get both to work with:
- Bokeh 1.4
- Panel 0.6.4
But any other version of Panel causes the Panel app to break. Even with the recent PR that updates compatibility with Bokeh 2.0, I'm still getting the error:

Environment (```conda list```):

<details>

```
(tethys-bokeh2)$ conda list
# packages in environment at tethys-bokeh2:
#
# Name                    Version                   Build  Channel
appdirs                   1.4.3                      py_1    conda-forge
asgiref                   3.2.5                      py_0    conda-forge
asn1crypto                1.3.0                    py37_0    conda-forge
asynctest                 0.12.2                     py_0    conda-forge
attrs                     19.3.0                     py_0    conda-forge
autobahn                  20.2.2                     py_0    conda-forge
automat                   20.2.0                     py_0    conda-forge
bcrypt                    3.1.7            py37h0b31af3_0    conda-forge
beautifulsoup4            4.8.2                    py37_0    conda-forge
blinker                   1.4                        py_1    conda-forge
bokeh                     2.0.0            py37hc8dfbb8_0    conda-forge
bumpversion               0.5.3                   py_1001    conda-forge
bzip2                     1.0.8                h0b31af3_2    conda-forge
ca-certificates           2019.11.28           hecc5488_0    conda-forge
certifi                   2019.11.28       py37hc8dfbb8_1    conda-forge
cffi                      1.14.0           py37h356ff06_0    conda-forge
cftime                    1.1.1            py37h10e2902_0    conda-forge
channels                  2.4.0                      py_0    tethysplatform/label/dev
chardet                   3.0.4           py37hc8dfbb8_1006    conda-forge
click                     7.1.1              pyh8c360ce_0    conda-forge
cloudpickle               1.3.0                      py_0    conda-forge
conda                     4.8.3            py37hc8dfbb8_1    conda-forge
conda-package-handling    1.6.0            py37h9bfed18_2    conda-forge
condorpy                  0.5.3                      py_0    tethysplatform/label/dev
constantly                15.1.0                     py_0    conda-forge
coverage                  5.0.4            py37h9bfed18_0    conda-forge
cryptography              2.8              py37he655712_2    conda-forge
curl                      7.68.0               h8754def_0    conda-forge
cytoolz                   0.10.1           py37h0b31af3_0    conda-forge
[...]

PR #546, opened 2020-04-10
Bokeh/Panel hotfix (#540)

msouff:
#540 fix

@sdc50 I confirm this also works with Bokeh 2.0.1


sdc50:
I completely missed that this happened. I've been MIA. Thanks @msouff!

Issue #545, opened 2020-04-10
Password Reset URLs without pattern name


ezrajrice:
Using the password recovery feature in Tethys 3.1.0 throws an error due to password_reset_done not having a valid pattern name.
Tethys Log (Production Install):
```
2020-04-10 09:36:08,091 ERROR    Internal Server Error: /accounts/password/reset/
Traceback (most recent call last):
  File "./django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "./django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "./django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "./django/views/generic/base.py", line 71, in view
    return self.dispatch(request, *args, **kwargs)
  File "./django/utils/decorators.py", line 45, in _wrapper
    return bound_method(*args, **kwargs)
  File "./django/utils/decorators.py", line 142, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "./django/contrib/auth/views.py", line 220, in dispatch
    return super().dispatch(*args, **kwargs)
  File "./django/views/generic/base.py", line 97, in dispatch
    return handler(request, *args, **kwargs)
  File "./django/views/generic/edit.py", line 142, in post
    return self.form_valid(form)
  File "./django/contrib/auth/views.py", line 234, in form_valid
    return super().form_valid(form)
  File "./django/views/generic/edit.py", line 57, in form_valid
    return HttpResponseRedirect(self.get_success_url())
  File "./django/views/generic/edit.py", line 51, in get_success_url
    if not self.success_url:
  File "./django/utils/functional.py", line 151, in __wrapper__
    res = func(*self.__args, **self.__kw)
  File "./django/urls/base.py", line 90, in reverse
    return iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs))
  File "./django/urls/resolvers.py", line 673, in _reverse_with_prefix
    raise NoReverseMatch(msg)
django.urls.exceptions.NoReverseMatch: Reverse for 'password_reset_done' not found. 'password_reset_done' is not a valid view function or pattern name.
```

ezrajrice:
[...]

PR #547, opened 2020-04-11
Broken URIs

ezrajrice:
Adds tests and fix for #545

PR #548, opened 2020-04-11
Broken URIs - hotfix

ezrajrice:
Added tests and fix for #545 - Password Reset URLs without pattern name.


ezrajrice:
@swainn This is ready to be merged once Travis finishes tests for PR #547

PR #568, opened 2020-06-28
Default username password for db command from portal_config.yml

swainn:
Use the DATABASES.default.USER and DATABASES.default.PASSWORD from the portal_config.yml as the default values for tethys db commands.

### DIFF
--- a/tethys_cli/db_commands.py
+++ b/tethys_cli/db_commands.py
@@ -42,9 +42,13 @@ def add_db_parser(subparsers):
     db_parser.add_argument('-d', '--database', dest='db_alias',
                            help="Name of the database options from portal_config.yml to use (e.g. 'default').")
     db_parser.add_argument('-n', '--username', dest='username',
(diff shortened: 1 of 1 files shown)


swainn:
I started this thinking there was a bug. I ran the `tethys db create` command without the `--username` and `--password` arguments thinking that it would get the values from the portal config. But when I ran the `tethys db migrate` I got an authentication error.

Turns out that the username and password are not derived from the portal_config.yml. So it used the default password 'pass'. However, the host and port are.

Seems that all the values should be derived from the portal_config.yml for consistency.

swainn:
@SarvaPulla @sdc50 after discussing with you the other day and after realizing that my changes don't actually do what I thought they would do, I've decided to withdraw this PR. I propose we keep it functioning how it is now. I've adjusted my course to reflect the correct way to call the `tethys db create` command accordingly by not assuming the username and password options will be populated from `portal_config.yml`.

PR #572, opened 2020-07-08
Collectstatic Copy App Static Dirs Instead of Link

swainn:
Changes:

- Changed the `tethys manage collectstatic` command to copy the static directory of apps to STATIC_ROOT by default instead of link.
- Can still use link option by providing the `--link` option: `tethys manage collectstatic --link`

Motivation: 

1. When installing on CentOS with SELinux enforcing, it is easiest to keep all static files in the web directory (`/var/www`) b/c it is already configured to allow access to files from HTTP applications. However, if the files are linked, to somewhere outside of that directory, then you'll hit the SELinux permissions wall.
2. All other static files from Tethys are copied, not linked, per the Django `collectstatic` command. This will make app static directories in STATIC_ROOT consistent with the other static directories.

### DIFF
--- a/tests/unit_tests/test_tethys_apps/test_management/test_commands/test_pre_collectstatic.py
+++ b/tests/unit_tests/test_tethys_apps/test_management/test_commands/test_pre_collectstatic.py
@@ -12,6 +12,18 @@ class ManagementCommandsPreCollectStaticTests(unittest.TestCase):
     def tearDown(self):
         pass
 
+    def test_add_arguments(self):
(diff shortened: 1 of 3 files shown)


swainn:
@SarvaPulla and @sdc50 I went ahead and merged this so it would be ready for my students to use on Monday. I considered our discussion at Tethys Scrum good enough for a review. All I've changed since then is updating the tests to match the new behavior. Let me know if you have any concerns with this.

PR #573, opened 2020-07-10
Settings

sdc50:
modify settings and docs to make them consistent


sdc50:
In this PR I only modify the `settings.py` file and the portal config docs. We had discussed modifying the settings CLI, but I decided that with the changes I made to the `settings.py` the additional benefit in modifying the CLI was small compared to the amount of code and maintenance it would require.

The documentation for the portal config still refers to several settings groupings (e.g. ``SESSION_CONFIG``). I ensured that all of these groupings are consistent with what is in the code. I also made sure that the name of the settings under the groupings are the same as the actual Django settings. Therefore both of the following are valid:

```bash
tethys settings -s SESSION_EXPIRE_AT_BROWSER_CLOSE True
```

```bash
tethys settings -s SESSION_CONFIG.SESSION_EXPIRE_AT_BROWSER_CLOSE True
```

Both will have the effect of setting the ``SESSION_EXPIRE_AT_BROWSER_CLOSE`` setting to ``True``. However they will appear differently in the ``portal_config.yml``

```yaml
SESSION_CONFIG:
   SESSION_EXPIRE_AT_BROWSER_CLOSE: True
```

vs.

```yaml
SESSION_EXPIRE_AT_BROWSER_CLOSE: True
```

This gives the user to maintain the organization of the supported groupings in the ``portal_config.yml`` while using the settings CLI, but it does not force them to. Either way any valid settings set through the settings command will be respected by the ``settings.py`` file in Django.

sdc50:
There are other potential groupings that we could add that may make as much sense as some of those that we already have. For example we could have a ``GRAVATAR_CONFIG``section. With the ``EMAIL_CONFIG`` the ``OAUTH_CONFIG`` and the ``ANALYTICS_CONFIG``, I'm not doing anything special with them. I just dump whatever settings they have into the ``settings.py`` file. So, in a similar fashion we could create any number of arbitrary groupings in the ``portal_config.yml`` to help organize things.

PR #577, opened 2020-07-31
Jobs Table Actions

sdc50:


sdc50:
image

sdc50:
@swainn I've modified the `get_logs` method for the `DaskJob` and the `CondorBase`. I don't have the correct environments to do integration tests for those cases, so if you or Tran are able to test them that would be helpful.

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
```diff
@@ -14,22 +14,75 @@
...
-    </td>
-  {% endif %}
 {% endif %}
+{% if show_actions %}
+  <td>
+    <div class="btn-group">
+      <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+        Actions <span class="caret"></span>
+      </button>
+      <ul class="dropdown-menu">
```
swainn, line 54: Can we right-align the dropdown menu with the button? Since it is wider than the button and in the far right column, it is possible that it will extend past the table when it is open.

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
```diff
@@ -14,22 +14,75 @@
...
-    <td>
-      <a class='btn btn-info btn-xs btn-job-show-log' href="javascript:void(0)" data-job-id="{{ job_id }}" data-toggle="modal" data-target="#modal-dialog-jobs-table-show-log">Logs</a>
-    </td>
-  {% endif %}
 {% endif %}
+{% if show_actions %}
+  <td>
+    <div class="btn-group">
+      <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+        Actions <span class="caret"></span>
```
swainn, line 52: We talked about this being the vertical triple dot icon instead of "Actions". What do you think about that?

tethys_compute/models/condor/condor_base.py
```diff
@@ -111,12 +116,31 @@ def _log_files(self):
...
+            else:
+                # Parse out logs for each job.
+                contents[log_file_type] = dict()
+                for job_log_type, job_log_path in log_file_path.items():
+                    contents[log_file_type][job_log_type] = partial(self.get_log_content, job_log_path)
+
+        return contents
+
+    def get_log_content(self, log_file):
+        content, _ = self.condor_object.execute(['cat', log_file])
```
swainn, line 142: Retrieving the logs in AGWA are failing with this error: 

```bash
tethys.tethys_gizmos.views.jobs_table:The following error occurred when retrieving log content for log 375 in job post_detention_basin_scenario error: 'Workflow' object has no attribute 'execute'
```
I don't use the base condor jobs so I don't know if this works for them, but it seems this won't work with the condor_object is a workflow?

swainn (changes requested):
I tested this fairly extensively with HTCondor workflows in AGWA. The Terminate and Delete actions seem to function properly. The Resubmit action works, but it doesn't reset the status of the Workflow to submitted, so the jobs table doesn't update automatically. I wonder if this needs to be handled at the TethysJob level? View logs did not work. See comment in review.

tethys_gizmos/views/gizmos/jobs_table.py
```diff
@@ -56,36 +71,96 @@ def show_log(request, job_id):
...
+            if isinstance(v, dict):
+                context['log_options'][f'log_select_{k}'] = SelectInput(
+                    name=f'log_{k}',
+                    multiple=False,
+                    options=[(key, key) for key in v.keys()],
+                )
+
+        html = render_to_string('tethys_gizmos/gizmos/job_logs.html', context)
+
+        def remove_callables(item):
```
swainn, line 95: Should remove_callable be handled  within the call to get_logs? So that get_logs return something fairly simple to parse without a lot of post processing that requires knowledge of how it is working underneath? Maybe I’m not understanding everything that is going on here.

sdc50: The `remove_callables` function handles some post processing so that it can be passed to JSON. In most cases when calling the `get_logs` method you want the callables to be there so that you can call them to get the log content.

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
sdc50: I'll try it out and send a screenshot.

tethys_compute/models/condor/condor_base.py
sdc50: I'll look closer at this to see if I can figure out what is going on.

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
sdc50: Good point. I'll see what I can do.

tethys_gizmos/urls.py
```diff
@@ -15,15 +15,17 @@
...
     url(r'^(?P<job_id>[\d.@+-]+)/resubmit$', jobs_table_views.resubmit, name='resubmit_job'),
     url(r'^(?P<job_id>[\d.@+-]+)/show-log', jobs_table_views.show_log, name='show_log'),
+    url(r'^(?P<job_id>[\d.@+-]+)/log-content/(?P<key1>[\w+_-]+)(?:/(?P<key2>.*))?', jobs_table_views.get_log_content,
+        name='log_content'),
     url(r'^(?P<job_id>[\d.@+-]+)/execute', jobs_table_views.execute, name='execute_job'),
     url(r'^(?P<job_id>[\d.@+-]+)/update-row', jobs_table_views.update_row, name='update_job_row'),
     url(r'^(?P<job_id>[\d.@+-]+)/update-workflow-nodes-row', jobs_table_views.update_workflow_nodes_row,
         name='update_workflow_nodes_row'),
     url(r'^(?P<job_id>[\d.@+-]+)/(?P<type>[\w-]+)/insert-bokeh-row', jobs_table_views.bokeh_row, name='bokeh_row'),
-    url(r'^(?P<job_id>[\d.@+-]+)/update-status', jobs_table_views.update_status, name='update_job_status'),
```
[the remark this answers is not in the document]
sdc50: Right. There were several methods handling update-status, but it wasn't being used. I think I went back and forth early on because I couldn't decide which approach to do, but I think we're pretty firmly in the `update-row` camp now, so I've removed all of the unused code.

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
sdc50: image

tethys_gizmos/static/tethys_gizmos/css/jobs_table.css
```diff
@@ -1,3 +1,15 @@
+.jobs-table, .dataTables_wrapper{
+    margin-bottom: 250px;
+}
+
+.jobs-table th{
+    white-space: nowrap;
+}
+
+.jobs-table td{
+    text-align: center;
```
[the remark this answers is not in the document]
sdc50: image
This is with
```css
.jobs-table th{
    white-space: nowrap;
    text-align: center;
}

.jobs-table td{
    text-align: center;
}
```

What are your thoughts on doing that?

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
sdc50: > Resubmit action works, but it doesn't reset the status of the Workflow to submitted, so the jobs table doesn't update automatically. I wonder if this needs to be handled at the TethysJob level?

This is handled at the `TethysJob` level under the `execute` method. I'm assuming that all jobs implementation of `_resubmit` will include a call to `execute`. In the case of `condor_base` it was calling `_execute` instead of `execute`, so I've fixed that which should also fix the issue of the status not updating.

tethys_gizmos/static/tethys_gizmos/css/jobs_table.css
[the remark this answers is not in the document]
swainn: My preference is for left-justfied. I think it ends up looking better 9 times out of 10. But this is your baby and the all center doesn't look terrible. I can always override it in my apps if it really bugs me.

The horizontal dots with the drop down arrow looks funny. Could we look at vertical dots and hide the dropdown arrow https://icons.getbootstrap.com/icons/three-dots-vertical/)?

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
swainn: Want to push the commit with the change so I can test?

tethys_compute/models/condor/condor_base.py
sdc50, line 142:
```suggestion
        content, _ = self.condor_object._execute(['cat', log_file])
```

sdc50: If only I were a little more familiar with that `condorpy` library. Then I wouldn't make little mistakes like this.

sdc50:
I remembered another feature that I wanted to add in. It allows status information to be presented in the jobs table:

image

I've implemented this so that if there is status information for the job, then the status indicator dot changes to an info icon and hovering over it displays the status message. Thoughts?

tethys_gizmos/static/tethys_gizmos/css/jobs_table.css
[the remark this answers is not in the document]
sdc50: I don't love the left-justified when you have a short number with a long header: 

image

But I agree with you that it's probably usually better left-justified and should be overridden in these cases.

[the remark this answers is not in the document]
sdc50: I think the three vertical dots look fine, but I might prefer just the dropdown arrow. 

image

tethys_compute/models/condor/condor_base.py
sdc50: This was the opposite as above. I called `execute` when I should have called `_execute`.

tethys_gizmos/static/tethys_gizmos/css/jobs_table.css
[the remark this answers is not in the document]
swainn: I like just the dropdown arrow.

tethys_compute/models/condor/condor_base.py
swainn: The logs are working in HTCondor. I'm a big fan of this feature.

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
swainn: The status column now starts updating after resubmitting. However, the detailed tree view does not:

image

 I think we just need to reinitiate the JavaScript loop to get it going again.

swainn (changes requested):
Added one response to an open issue. The Resubmit now resets the status column so that it updates, but the detailed tree view is stuck showing everything completed. I think we just need to reinitiate it like we do with the status column.

tethys_gizmos/static/tethys_gizmos/js/jobs_table.js
```diff
@@ -81,43 +79,79 @@ function bind_refresh_button(btn){
...
 function bind_resubmit_button(btn){
-    btn = $(btn);
     var job_id = $(btn).data('job-id');
     $(btn).on('click', function(){
         $("#jobs_table_overlay").removeClass('hidden');
         var resubmit_url = '/developer/gizmos/ajax/' + job_id + '/resubmit';
         $.ajax({
             url: resubmit_url
         }).done(function(json){
+            update_row($(btn).closest('tr'));
```
swainn, line 154:
```suggestion
            update_row($(btn).closest('tr'));
            update_workflow_nodes_row($(btn).closest('tr').next('tr'));
```

swainn: ^ This should do the trick to get the tree view updating again.

tethys_gizmos/static/tethys_gizmos/css/jobs_table.css
```diff
@@ -1,3 +1,11 @@
+.jobs-table, .dataTables_wrapper{
+    margin-bottom: 250px;
+}
+
+.jobs-table th{
+    white-space: nowrap;
+}
+
```
swainn, line 8:
```suggestion
.jobs-table>tbody>tr>td {
    vertical-align: middle;
}
```

Before:
image


After:
image

tethys_gizmos/templates/tethys_gizmos/gizmos/job_row.html
sdc50: Committed your suggestion

swainn (approved):
Excellent work on this. Thanks for the quick turn around.

PR #580, opened 2020-08-06
App access setting

msouff:
Added a setting configuration to restrict app access based on user permissions. The functionality is now off by default.


sdc50:
@swainn I approve of this PR, but you mentioned that you wanted to review it too.

tethys_apps/views.py
```diff
@@ -39,7 +40,7 @@ def library(request):
                 configured_apps.append(app)
             else:
                 unconfigured_apps.append(app)
-        elif user_can_access_app(request.user, app):
+        elif user_can_access_app(request.user, app) or not getattr(settings, "ENABLE_RESTRICTED_APP_ACCESS", False):
```
swainn, line 43: The check on `ENABLE_RESTRICTED_APP_ACCESS` should be in `user_can_access_app`, no? Is there a reason we wouldn't put it in there?

Also, this only hides the app tile. What if the user enters the URL manually? Isn't there a middleware where we need to check for this?

msouff:
I moved the restricted access check to the user_can_access_app function. This also takes care of the middleware part.

PR #595, opened 2020-09-10
hotfix for docker client when docker daemon is not running

sdc50:


swainn:
@sdc50 let's enable the osx testing again to verify that the fix is working as we discussed in scrum.

swainn (approved):

PR #596, opened 2020-09-10
Adds Lockout Capabilities

swainn:
image
- Adds automatic account locking
- Implemented with [Django Axes](https://django-axes.readthedocs.io/en/latest/)
- Prevent brute force login hacks
- The default behavior is to lockout after 3 failed login attempts with a half-hour cool off time before the user can log in (in production).
- Not enabled when `DEBUG` is `True`
- All new dependencies introduced are packaged on Conda Forge


swainn:
@sdc50 and @SarvaPulla this one is ready for review as well. This has been rebased with the master. Unit tests and conda build are passing, but docker build continues to cause the failure of the Travis check.

swainn:
@ezrajrice this is ready for you're review. This branch has been rebased with the other features in it too.

tethys_portal/templates/tethys_portal/accounts/lockout.html
```diff
@@ -0,0 +1,44 @@
...
+          {% endblock %}
+        </div>
+      </div>
+      <div class="col-lg-6">
+        <div class="row">
+          <div class="col-lg-12">
+            <div class="lockout-message-container">
+              <p>You have exceeded the amount of allowed login attempts ({{ failure_limit }}).</p>
+              {% if cooloff_time %}
+                <p>Your account will be unlocked {{ cooloff_time|human_duration }}, after which you may try again.</p>
```
sdc50, line 34: This is nit-picky, but it would make more sense to me to say "Your account will be _locked for_ 30 min, after which you may try again."

I'm assuming the humanize thing makes it say _in 30 min_, which is fine, but in that case I wonder if it would sound better to say: "Your account will be unlocked in 30 min, _at which point you may try again._" 

If you disagree then go ahead and merge as is.

sdc50 (commented):
I approve, but left one suggestion.

tethys_portal/templates/tethys_portal/accounts/lockout.html
swainn: @sdc50 thanks for the suggestion. I modified it to match your second suggestion because as you surmised, humanize returns "in 30 minutes".

sdc50 (approved):

PR #604, opened 2020-10-09
added the ability to pass ENV and VIEWPARAMS to WMS layer

msouff:


swainn (approved):
Looks good. Be nice to support these features of MVLayer in the Cesium Map View

Issue #592, opened 2020-09-01
Management Form data missing or tampered bug


ezrajrice:
The following error occurs when trying to create a new (second) Tethys user account after successfully creating one. The username and password fields are unique on the second attempted account creation. After this error has occurred, I am unable to create any other user account until I restart Tethys.
OS: Ubuntu 18.04 LTS
```
 File "./django/utils/decorators.py", line 142, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "./django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "./django/contrib/admin/sites.py", line 223, in inner
    return view(request, *args, **kwargs)
  File "./django/utils/decorators.py", line 45, in _wrapper
    return bound_method(*args, **kwargs)
  File "./django/views/decorators/debug.py", line 76, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
  File "./django/utils/decorators.py", line 45, in _wrapper
    return bound_method(*args, **kwargs)
  File "./django/utils/decorators.py", line 142, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "./django/contrib/auth/admin.py", line 99, in add_view
    return self._add_view(request, form_url, extra_context)
  File "./django/contrib/auth/admin.py", line 126, in _add_view
    return super().add_view(request, form_url, extra_context)
  File "./django/contrib/admin/options.py", line 1645, in add_view
    return self.changeform_view(request, None, form_url, extra_context)
  File "./django/utils/decorators.py", line 45, in _wrapper
    return bound_method(*args, **kwargs)
  File "./django/utils/decorators.py", line 142, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "./django/contrib/admin/options.py", line 1529, in changeform_view
    return self._changeform_view(request, object_id, form_url, extra_context)
  File "./django/contrib/admin/options.py", line 1571, in _changeform_view
    if all_valid(formsets) and form_validated:
  File "./django/forms/formsets.py", line 448, in all_valid
    valid &= formset.is_valid()
  File "./django/forms/formsets.py", line 301, in is_valid
    self.errors
  File "./django/forms/formsets.py", line 281, in errors
    self.full_clean()
  File "./django/forms/formsets.py", line 322, in full_clean
    for i in range(0, self.total_form_count()):
  File "./django/forms/formsets.py", line 110, in total_form_count
[...]

PR #606, opened 2020-10-22
Fix for Issue 592

swainn:
The issue was caused by how we added the custom Quota's fields to the change/edit User admin form:

```python
class CustomUser(UserAdmin):
    def change_view(self, *args, **kwargs):
        self.inlines = [UserQuotasSettingInline]
        return super().change_view(*args, **kwargs)
```

Changing `self.inlines` for the `change_view` of the `UserAdmin` form permanently changed it for all views including the `add_view`, which would cause the error. I'm guessing this is because `self.inlines` is a class property.

The fix was to remove the `UserQuotasSettingInline` for the `add_view`:

```python
class CustomUser(UserAdmin):
    def add_view(self, *args, **kwargs):
        if UserQuotasSettingInline in self.inlines:
            self.inlines.pop(self.inlines.index(UserQuotasSettingInline))
        return super().add_view(*args, **kwargs)

    def change_view(self, *args, **kwargs):
        if UserQuotasSettingInline not in self.inlines:
            self.inlines.append(UserQuotasSettingInline)
        return super().change_view(*args, **kwargs)
```

### DIFF
--- a/tests/unit_tests/test_tethys_apps/test_admin.py
+++ b/tests/unit_tests/test_tethys_apps/test_admin.py
@@ -254,6 +254,17 @@ class TestTethysAppAdmin(unittest.TestCase):
         mock_ua_add_view.assert_called()
         self.assertNotIn(UserQuotasSettingInline, ret.inlines)
 
+        # Repeat to complete full cycle (change -> add -> change -> add)
+        # Add custom inline when change_view is called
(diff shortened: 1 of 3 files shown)


swainn:
Fixes issue #592

sdc50 (approved):

PR #608, opened 2020-10-26
Asgi conf update

rfun:
On CentOS Systems, I have noticed that `sudo supervisorctl restart all` fails with an RPC error. The reason this happens is that based on the configuration file, supervisord doesn't kill the previous process (why this doesn't happen in other linux systems, I don't know. But I have seen this specifically in CentOS 6). However, adding the correct lines for the kill command, it works and restarts fine in centos systems as well. 

Here is a thread that describes the issue https://github.com/Supervisor/supervisor/issues/600

I have updated tethys's asgi template to ensure it produces a config file that works flawlessly for centos systems as well.


swainn (commented):
@rfun have you verified that this works in other Linux distros as well? Debian/Ubuntu-based systems are the others that are used by our users in production.

rfun:
> @rfun have you verified that this works in other Linux distros as well? Debian/Ubuntu-based systems are the others that are used by our users in production.

I just tested it on Tethys.byu.edu and it worked in Ubuntu. I don't have access to a debian production server of tethys. Sorry! 
I could possibly set one up on aws just to test it out on debian......

swainn (approved):

PR #625, opened 2020-12-09
Custom settings updates

### LINKED ISSUE #622: Set Custom Setting [feature request]
Tethys apps have a get_custom_setting that returns the value assigned to the custom setting. As of right now that value has to be assigned using the admin or the CustomSetting database object.

I propose we add a "set_custom_setting" method to the TethysAppBase class.

### LINKED ISSUE #623: Custom Setting with Type UUID
Tethys app custom settings support string, integer, float, and boolean. Right now we have to use TYPE_STRING for apps that require a unique identifier setting due to the lack of support for TYPE_UUID. This can can easily result in errors since in reality the these settings would not accept any string but a UUID.

Postgres does support UUID columns so it would be a matter of adding it as an extra option to the Tethys custom setting class (CustomSetting).

msouff:
addresses the custom setting issues.

Resolves #622, Resolves #623


tethys_apps/base/app_base.py
```diff
@@ -924,6 +925,53 @@ def get_custom_setting(cls, name):
...
+        if custom_setting.type == 'STRING':
+            type_matches = isinstance(value, str)
+        elif custom_setting.type == 'INTEGER':
+            type_matches = isinstance(value, int)
+        elif custom_setting.type == 'FLOAT':
+            type_matches = isinstance(value, float)
+        elif custom_setting.type == 'BOOLEAN':
+            type_matches = str(value).lower() in ['true', 'false', 'yes', 'no', 't', 'f', 'y', 'n', '1', '0']
+        elif custom_setting.type == 'UUID':
+            type_matches = isinstance(value, uuid.UUID)
```
swainn, line 967: Technically they could provide a string version of a UUID that I would think we count as valid:

```
>>> str(uuid.uuid4())
'f50ec0b7-f960-400d-91f0-c42a6d44e3d0'
```

In addition to checking if it is an instance of a `uuid.UUID` I suggest you try to instantiate a `uuid.UUID`  from the given value.

swainn (changes requested):
One suggestion. See comments.

tethys_apps/base/app_base.py
msouff: You are right. I should have caught that since I did it something similar with the boolean type.

tests/unit_tests/test_tethys_apps/test_base/test_app_base.py
```diff
@@ -648,12 +648,12 @@ def test_set_custom_setting_object_not_exist(self, mock_app, mock_tas_dne):
     @mock.patch('tethys_apps.models.TethysApp')
     def test_set_custom_setting_type_not_match(self, mock_app):
         setting_name = 'fake_setting'
-        mock_app.objects.get().custom_settings.get.return_value = mock.MagicMock(type='STRING')
+        mock_app.objects.get().custom_settings.get.return_value = mock.MagicMock(type='UUID')
```
swainn, line 651: Please add a new test to test the UUID type_match functionality instead of repurposing an existing test. I'd like to get away from doing the minimal to reach 100% coverage.

swainn (changes requested):

tests/unit_tests/test_tethys_apps/test_base/test_app_base.py
msouff: https://github.com/msouff/tethys/blob/custom_settings_updates/tethys_apps/base/app_base.py#L957-L976

I picked the STRING type previously because it was the first in order, but what actually needs to be tested here is that the 'type_matches' variable is False and that the exception in line 976 is raised. I thought changing the type to UUID was a clever way to test this while hitting line 970 in the process since we still want to hit every line.

Testing the exception at line 970 separately doesn't make sense because it is just passing. The separate test would be unrealistic (like, I'd side_effect the exception and then test that it is throwing that exception?). Let me know what you think.

tethys_apps/base/app_base.py
```diff
@@ -924,6 +925,56 @@ def get_custom_setting(cls, name):
...
+    def set_custom_setting(cls, name, value):
+        """
+        Assign the value of a CustomSetting for the app.
+
+        Args:
+            name(str): The name of the CustomSetting as defined in the app.py.
+            value(str/int/float/boolean/uuid.UUID): the value of the customSetting.
+
+        Returns:
+            variable: Value of the CustomSetting or None if no value assigned.
```
swainn, line 938: Remove. Doesn't return anything.

swainn (approved):

Issue #626, opened 2020-12-10
MFA bug is hiding API Token


ezrajrice:
I’m having an issue where the API token isn’t visible for users when MFA is configured and not required. 
My sample account has Email MFA configured and is working, but cannot see the API Token.
 
My MFA configuration in the portal_config.yml file is:
```YAML
  MFA_CONFIG:
    MFA_REQUIRED: False
    ADMIN_MFA_REQUIRED: False
    SSO_MFA_REQUIRED: False
    MFA_RECHECK: True
    MFA_RECHECK_MIN: 60
    MFA_RECHECK_MAX: 180
    MFA_QUICKLOGIN: True
    TOKEN_ISSUER_NAME: 'SMARTLINE Portal (Dev)'
    MFA_UNALLOWED_METHODS: ['U2F', 'FIDO2', 'Trusted_Devices']
```
I believe the bug to be in the template in tethys_portal/templates/tethys_portal/user/profile.html
```html
              <span class="parameter">Token:</span>
              {% if user.username == context_user.username %}
                {% if mfa_required and has_mfa %}
                <span class="value">{{ user_token }}</span>
                {% else %}
                <span class="value">Enable 2-Step verification to view token</span>
                {% endif %}
              {% else %}
                <span class="label label-info">Private</span>
              {% endif %}
```
This is the change that I think should be made:
`{% if mfa_required and has_mfa %}` -> `{% if not mfa_required or (mfa_required and has_mfa) %}`

The issue seems to only occur when MFA_REQUIRED is False.

swainn:
@ezrajrice I have confirmed the issue. I give your suggestion a try and get a hotfix PR going.

swainn:
@ezrajrice I submitted the PRs with the fix for this. Please take a look and confirm that it will address your problem.

ezrajrice:
FYI for future reference, from django's [documentation](https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#if):

> Use of both **and** and **or** clauses within the same tag is allowed, with **and** having higher precedence than **or** e.g.:
> ```django
> {% if athlete_list and coach_list or cheerleader_list %}
> ```
> will be interpreted like:
> ```python
> if (athlete_list and coach_list) or cheerleader_list
> ```
> Use of actual parentheses in the **if** tag is invalid syntax. If you need them to indicate precedence, you should use nested **if** tags.

swainn:
[...]

PR #627, opened 2020-12-10
Fixes Issue #626

swainn:
I had to implement the logic check in the controller, rather than the template as @ezrajrice suggested b/c the templating syntax doesn't support parenthesis. I also added four new test cases to cover all combinations of had_mfa and mfa_required so we can catch this issue in the future if it occurs.

### DIFF
--- a/tests/unit_tests/test_tethys_portal/test_views/test_user.py
+++ b/tests/unit_tests/test_tethys_portal/test_views/test_user.py
@@ -201,7 +201,7 @@ class TethysPortalUserTests(unittest.TestCase):
             'quota': None,
             'has_mfa': True,
             'mfa_required': False,
-            'show_user_token_mfa': True  # Show user token b/c mfa is required and user has setup mfa
+            'show_user_token_mfa': True  # Show user token b/c not mfa is required
         }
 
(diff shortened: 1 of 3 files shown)


ezrajrice:
Looks good! Thanks!

tests/unit_tests/test_tethys_portal/test_views/test_user.py
```diff
@@ -72,6 +97,117 @@ def test_profile(self, mock_get_user, mock_token_get_create, mock_render, mock_g
...
+        mock_has_mfa.return_value = True
+
+        expected_context = {
+            'context_user': mock_context_user,
+            'user_token': mock_user_token.key,
+            'current_use': '0 bytes',
+            'quota': None,
+            'has_mfa': True,
+            'mfa_required': False,
+            'show_user_token_mfa': True  # Show user token b/c mfa is required and user has setup mfa
```
msouff, line 204: Update the comment to "Show user token b/c not mfa_required" or something similar

swainn: Done.

msouff (approved):

Issue #630, opened 2021-01-13
Static Images with Metadata


ezrajrice:
Is there a reason to keep the metadata on the static files (PDF, JPG/JPEG, TIFF, GIF, PNG, PSD, BMP, RAW, CR2, CRW, PICT, XMP, DNG)?

These are getting flagged in penetration testing as "Metadata Information Leakage" with a CVSS Score of 2.6 (low risk). I've looked at the metadata and found that if we just make sure that we remove the metadata formats such as EXIF, IPTC, and XMP, then this shouldn't get flagged by most penetration testing software.

Example finding:
Request URL http://localhost:8000/static/tethys_portal/images/tethys-logo-25.png
Metadata found:
```json
{
  "PNG-IHDR": {
    "Image Width": "25",
    "Image Height": "25",
    "Bits Per Sample": "8",
    "Color Type": "True Color with Alpha",
    "Compression Type": "Deflate",
    "Filter Method": "Adaptive",
    "Interlace Method": "No Interlace"
  },
  "PNG-pHYs": {
    "Pixels Per Unit X": "11811",
    "Pixels Per Unit Y": "11811",
    "Unit Specifier": "Metres"
  },
  "XMP": {
    "XMP Value Count": "28",
    "Create Date": "2015-02-05T23:38:28-07:00",
    "Creator Tool": "Adobe Photoshop CC 2014 (Macintosh)",
    "Metadata Date": "2015-03-13T16:19:16-06:00",
    "Modify Date": "2015-03-13T16:19:16-06:00"
  },
  "PNG Chromaticities": {
    "White Point X": "31269",
    "White Point Y": "32899",
    "Red X": "63999",
    "Red Y": "33001",
    "Green X": "30000",
    "Green Y": "60000",
    "Blue X": "15000",
    "Blue Y": "5999"
  }
}
```

swainn:
@ezrajrice I was actually unaware of the metadata on the images. I'm guessing it would not be a problem to remove that. Is there a good tool that can do this in batches?

ezrajrice:
I've used [exiftool](https://exiftool.org/) on Ubuntu. It's supported on Windows, MacOS, and various flavors of Linux. I just found that the Linux install was smoother to install and use.

Installation on Ubuntu, Debian, Mint, Kali:  

> `$ sudo apt install libimage-exiftool-perl`

Installation on Fedora, CentOS, RedHat:  

> `$ sudo dnf install perl-Image-ExifTool.noarch`

[...]

PR #648, opened 2021-02-08
Fix for "Metadata Information Leakage". Removed EXIF, IPTC, and XMP m…

ezrajrice:
Fix for "Metadata Information Leakage". Removed EXIF, IPTC, and XMP metadata from png and jpg files to address Issue #630


swainn (approved):

swainn:
@ezrajrice just to verify, this has no impact on the rendering of the images?

ezrajrice:
> 
> 
> @ezrajrice just to verify, this has no impact on the rendering of the images?

That's correct. As far as I was able to tell this didn't affect the display of any image I loaded. There are some one off cases where it might affect loading. A good explanation of those cases can be found [here](https://dexecure.com/blog/impact-of-metadata-on-image-performance/). I wasn't able to detect any differences when I was looking at them.

PR #662, opened 2021-02-25
Simplify/secure CSRF token ajax handling

sdc50:
This moves the retrieving of the CSRF token from an AJAX call to having be embedded in the page in a script tag. It also moves code from multiple places into the app_base.js file, so it's only in one place.

### DIFF
--- a/tethys_apps/static/tethys_apps/js/app_base.js
+++ b/tethys_apps/static/tethys_apps/js/app_base.js
@@ -299,3 +299,25 @@ var TETHYS_APP_BASE = (function() {
 }()); // End of package wrapper
 // NOTE: that the call operator (open-closed parenthesis) is used to invoke the library wrapper
 // function immediately after being parsed, returning the public interface object.
+
+
+// Global Functions for App Util
+
+/*****************************************************************************
+ *
+ * Cross Site Request Forgery Token Configuration
+ *   copied from (https://docs.djangoproject.com/en/1.7/ref/contrib/csrf/)
+ *
+ *****************************************************************************/
+
+function csrfSafeMethod(method) {
+    // these HTTP methods do not require CSRF protection
+    return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
+}
+$.ajaxSetup({
+    beforeSend: function(xhr, settings) {
+        if (!csrfSafeMethod(settings.type) && !this.crossDomain) {
+            xhr.setRequestHeader("X-CSRFToken", csrftoken);  // Note csrftoken is defined in app_base.html
+        }
+    }
+});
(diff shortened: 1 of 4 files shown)


swainn:
@sdc50 thanks for the contribution. I'm generally supportive, but I have a few questions for my understanding:

1. Did you mean instead of retrieving it from a cookie? I don't think we were ever retrieving the CSRF token from an AJAX request.
2. What is the primary concern that is being addressed here? Is it less secure to retrieve it from the cookie?
3. The CSRF token is always embedded in the page when it is used. It seems the main difference here is that the token will be embedded on every page and is then automatically added to every XHR request. The result is developers won't have to do it themselves when submitting POST requests via AJAX. Is that correct?

tethys_apps/static/tethys_apps/js/app_base.js
```diff
@@ -298,4 +298,26 @@ var TETHYS_APP_BASE = (function() {
...
+// Global Functions for App Util
+
+/*****************************************************************************
+ *
+ * Cross Site Request Forgery Token Configuration
+ *   copied from (https://docs.djangoproject.com/en/1.7/ref/contrib/csrf/)
+ *
+ *****************************************************************************/
+
+function csrfSafeMethod(method) {
```
swainn, line 313: Make this a private method of `TETHYS_APP_BASE`. I don't think this has general use.

tethys_apps/static/tethys_apps/js/app_base.js
```diff
@@ -298,4 +298,26 @@ var TETHYS_APP_BASE = (function() {
...
+    // these HTTP methods do not require CSRF protection
+    return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
+}
+$.ajaxSetup({
```
swainn, line 317: Run this in the page init logic in `TETHYS_APP_BASE`.

tethys_apps/templates/tethys_apps/app_base.html
```diff
@@ -266,6 +266,10 @@
     {% endcomment %}
 
     {% block scripts %}
+      {% csrf_token %}
+      <script>
```
swainn, line 270: Move the definition of the `csrftoken` variable to the `app_base.js`

swainn (changes requested):

tethys_apps/static/tethys_apps/js/app_base.js
```diff
@@ -196,6 +196,11 @@ var TETHYS_APP_BASE = (function() {
         $(app_content_selector).removeClass('show-app-content');
    };
 
+    csrf_safe_method = function(method) {
+        // these HTTP methods do not require CSRF protection
+        return (/^(GET|HEAD|OPTIONS|TRACE)$/.test(method));
+    }
```
swainn, line 202: Need a `;` at the end of line 202.

tethys_apps/templates/tethys_apps/app_base.html
```diff
@@ -254,6 +254,8 @@
       {% gizmo_dependencies css %}
     {% endblock %}
 
+    {% csrf_token %}
```
swainn, line 257: Place this in a block so it can be overridden.

swainn (changes requested):
Just a few more minor suggestions.

swainn (approved):

PR #663, opened 2021-02-25
Restrict Open User Profiles

sdc50:
Provide a settings option to turn off having open user profiles. If this settings is `False` then when a user tries to go to the URL of another user they will be redirected to their own profile.

image

This provides additional security because it prevents users from being able to figure out other users usernames.

### DIFF
--- a/tests/unit_tests/test_tethys_portal/test_views/test_user.py
+++ b/tests/unit_tests/test_tethys_portal/test_views/test_user.py
@@ -19,6 +19,7 @@ class TethysPortalUserTests(unittest.TestCase):
     def tearDown(self):
         pass
 
+    @override_settings(OPEN_USER_PROFILES=True)
     @override_settings(MFA_REQUIRED=False)
     @mock.patch('tethys_quotas.utilities.log')
     @mock.patch('tethys_portal.views.user.has_mfa')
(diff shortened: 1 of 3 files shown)


swainn:
@sdc50 I believe @ezrajrice had similar concerns with the user profiles and was working on a solution. I believe his proposal was to take it a step further and remove the username from the URL and change the behavior to always show the logged-in user's profile. This would negate these changes b/c there would be no other users' profiles.

I was supportive of @ezrajrice approach b/c I don't think the user profile feature is used and it creates opportunities for security leaks of user information and settings if not handled properly.

That said, I don't know what @ezrajrice status or timeframe is on implementing that solution. I would hate to merge this to have it ripped out in a month or so.

ezrajrice:
@swainn I was planning on getting this done this weekend or next weekend.  Personally, I'd prefer to remove the username altogether because browsers frequently cache URLs. Correct me if I'm wrong, but my understanding is that the never_cache flag just makes the browser not cache the page itself and doesn't prevent it caching the URL pattern.

swainn:
@sdc50 would that solution accomplish what you are needing?

sdc50:
@ezrajrice yes I like that idea. I just took a first pass at it, so I'll push my branch if you want to use that as a baseline.

sdc50:
replaced by #664

PR #664, opened 2021-02-25
remove username from urls

sdc50:


ezrajrice (approved):
@sdc50 It all looks good to me. I guess this means I can delete my local branch for this task :)

tethys_portal/urls.py
```diff
@@ -96,7 +96,7 @@
     url(r'^accounts/', include((account_urls, 'accounts'), namespace='accounts')),
     url(r'^captcha/', include('captcha.urls')),
     url(r'^oauth2/', include((oauth2_urls, psa_urls.app_name), namespace='social')),
-    url(r'^user/(?P<username>[\w.@+-]+)/', include((user_urls, 'user'), namespace='user')),
```
swainn, line 99: I think for backwards compatibility, we ought to retain a URL pattern with the username in it that maps to a simple controller that redirects to the URL without the username. 

In that controller we should also use the Django messages API to display a message to the user indicating that the URL with usernames in it will not be supported in future versions.

Then we mark the url with username and controller as deprecated (at the very least with a comment) to be removed in version 4.0.

swainn (changes requested):
I'm very appreciative to see this change getting implemented. This is something I've been wanting to see done for a while. I have only one suggestion to provide some backward compatibility. See comment below.

swainn:
@ezrajrice I discussed this PR with @sdc50 today on Tethys Scrum. He mentioned that he hasn't been able to do much manual testing on these changes yet. Think you could pull the branch and test it out? We'd want to manually test all the views that are affected by this change.

tethys_portal/urls.py
swainn: @ezrajrice would this be ok with what you need ^

ezrajrice: @swainn What use cases are there that would require this kind of backwards compatibility? Are you concerned about breaking apps? Also, as long as the default behaviour is to use the URLs without the username, I'm fine with keeping this in with a deprecation warning.

ezrajrice:
> 
> 
> @ezrajrice I discussed this PR with @sdc50 today on Tethys Scrum. He mentioned that he hasn't been able to do much manual testing on these changes yet. Think you could pull the branch and test it out? We'd want to manually test all the views that are affected by this change.

I'll pull this PR and manually check each URL tomorrow. It's 11:15 PM over here so I don't expect I'll be awake for much longer.

tethys_portal/urls.py
swainn: @ezrajrice The only cases, I can think of would be if someone has the URL bookmarked or an app has the URL hard-coded through a link. This is a fairly low risk. But it also wouldn't take a lot for us to support the old URL and redirect to the new pattern. 

I guess my concern is: would retaining a URL pattern with a username in it still be flagged in your security testing? Though I don't know that it could be rightly called a "username" b/c we wouldn't use it at all. You could put "foo" in place of the username parameter and it would still have the same behavior as a real username.

ezrajrice: @swainn I see your point. After reviewing automated pen testing standards, I don't think this would get flagged because it automatically redirects to the URL without a username. Most pen testing just use web crawlers to find all of the URLs anyway so as long as this is the behaviour and there aren't any links in the views that include the username then I think this is fine.

swainn: > I think for backwards compatibility, we ought to retain a URL pattern with the username in it that maps to a simple controller that redirects to the URL without the username.
> 
> In that controller we should also use the Django messages API to display a message to the user indicating that the URL with usernames in it will not be supported in future versions.
> 
> Then we mark the url with username and controller as deprecated (at the very least with a comment) to be removed in version 4.0.

After looking closer at this, I've realized it won't be as simple as I thought to support backward compatible URLs. We'll have to duplicate all of the user url patterns and controllers... The effort does not justify the benefit in my mind.

swainn (approved):

swainn:
I tested all the endpoints affected by these changes with the exception of the clear workspaces endpoint for quotas. This included the social auth connect/disconnect endpoints. Looks good to me.

PR #666, opened 2021-03-03
Implements social login for ArcGIS (Online/Portal)

shawncrawley:
With these changes, ArcGIS Online or an ArcGIS Enterprise Portal can be  used to authenticate your Tethys Portal users. The portal_config.yml must be modified to include the proper entries in the `AUTHENTICATION_BACKENDS` and `OAUTH_CONFIGS` sections. 

For ArcGIS Online, the backend is `social_core.backends.arcgis` and the `OAUTH_CONFIG` settings are `SOCIAL_AUTH_ARCGIS_KEY` and `SOCIAL_AUTH_ARCGIS_SECRET`.

For ArcGIS Enterprise Portal, the backend is `tethys_services.backends.arcgis_portal.ArcGISPortalOAuth2` and the `OAUTH_CONFIG` settings are `SOCIAL_AUTH_ARCGIS_PORTAL_URL`, `SOCIAL_AUTH_ARCGIS_PORTAL_KEY`, and `SOCIAL_AUTH_ARCGIS_PORTAL_SECRET`. 

In either case, the key and secret are obtained from an "Application Item" that must be created and registered either on ArcGIS Online or the ArcGIS Enterprise Portal that will authenticate your users.


tests/unit_tests/test_tethys_services/test_backends/test_arcgis_portal.py
```diff
@@ -0,0 +1,101 @@
+from unittest import mock
+from django import test
+from tethys_services.backends.arcgis_portal import ArcGISPortalOAuth2
+
+
+@test.override_settings(
+    OAUTH_CONFIG={
+        'SOCIAL_AUTH_ARCGIS_PORTAL_URL': 'https://my.test.arcgis/portal'
+    }
+)
```
swainn, line 10: I took a look at why your tests are failing. I'm guessing that the `ArcGISPortalOAuth2` class and class level logic is processed before this decorator is applied, causing the error. Generally, logic at the class level causes issues when testing. Is there a method on the class that you can use to do your check for the portal url?

swainn (changes requested):
@shawncrawley great first pass. I have a few suggestions. See comments.

tests/unit_tests/test_tethys_services/test_backends/test_arcgis_portal.py
shawncrawley: The tests are now fixed, so you can give this one final review. Thanks!

swainn:
@shawncrawley thanks for getting this in. Although coveralls passed, there is one line that is still missing coverage. See: https://coveralls.io/builds/37822473/source?filename=tethys_services%2Fbackends%2Farcgis_portal.py#L27

Should be pretty easy to address. Just add another test where the portal URL has an ending `/`.

swainn (changes requested):
One minor change in your tests. I really like the `property` approach. Looks good.

tests/unit_tests/test_tethys_services/test_backends/test_arcgis_portal.py
```diff
@@ -0,0 +1,108 @@
...
+class ArcGISPortalBackendTest(test.SimpleTestCase):
+    strategy = mock.MagicMock()
```
swainn, line 12: Somehow this comment got removed... anyway... move `strategy` into the `setUp()` method (as `self.strategy`) so it is reset between tests.

shawncrawley: Applied. Looks like test and coverage passed this time.

swainn (approved):

PR #673, opened 2021-03-14
I18n

msouff:


swainn:
@msouff Thank you for submitting this! I think this is a great start, but I think we'll need to provide a bit more for it to be effective most. Specifically:

* We need to build out cli commands to allow users to generate and build their own po files. Then refactor so that the translation tags are in the template where we use the `site_globals` variable. Doing this will allow administrators to provide translations for their own content.
* We need to support translation in apps. Most likely by having users generate and store translations in the app source code itself and having Tethys search apps for translations.

 I've created a new branch called i18n and change the PR to merge into that branch. I'd like to take a closer look at this before I merge.

PR #680, opened 2021-04-07
Pin Channels to 2.*

swainn:
BYU is experiencing issues where installing Dev builds installs Channels 3.*, which breaks Tethys. This PR pins the channels dependency to 2.* for now.


swainn:
@rfun I've submitted this PR to pin the channels dependency as you suggested.

sdc50 (approved):

PR #696, opened 2021-06-10
WIP: Custom job actions

sdc50:
Adds ability to add custom actions to jobs.

This is supported by dynamically modifying the `TethysJob` model class with a new method (the custom action).  The API is simple. You just add custom actions to the `actions` argument of the `JobsTable` gizmo options as a tuple:

```python
def custom(job):
    # do custom stuff with job

def custom_enabled(job):
    if (# code to determine if custom action should be enabled):
        return True

jobs_table = JobsTable(
     ...
        actions=[
            'logs', 'terminate', 'delete',
            ('Custom', custom, custom_enabled, TethysJob),
        ],
    )

```

The tuple should be in the form `(ACTION_NAME, CUSTOM_METHOD, [[ENABLE_METHOD], JOB_TYPE])`

Where `ENABLED_METHOD` and `JOB_TYPE` are optional.

As an alternative to providing a tuple you can also pass in a `CustomJobAction` object:

```python
        actions=[
            'logs', 'terminate', 'delete',
            CustomJobAction(label='Custom', callback=custom, enabled_callback=custom_enabled, job_type=TethysJob),
        ],
```

Custom Actions will show up in a new section at the bottom of the actions list:

image

For this screenshot this is the configuration I used:

```python
        actions=[
            'logs', 'terminate', 'delete',
            ('Resubmit', 'resubmit', resubmitable),
            ('Archive', 'archive', archivable, UitPlusJob),
            ('Restore', 'restore', is_archived, UitPlusJob),
        ],
```
Note that although `resubmit` is a standard action, I needed custom logic to enable/disable it. In this case rather than passing in a custom method for the `callback` I just passed a string with the name of method that the `job_type` class already has.


tethys_gizmos/static/tethys_gizmos/js/jobs_table.js
```diff
@@ -384,21 +415,28 @@ function update_row(table_elem){
...
 
+    var data = {
+        column_fields: column_fields,
+        show_status: show_status,
+        show_actions: show_actions,
+        monitor_url: monitor_url,
+        results_url: results_url,
+        actions: actions,
+        custom_actions: custom_actions
+    }
```
swainn, line 434: Missing semicolon

swainn (changes requested):
I'd like to discuss more with you, but overall, I think this is a good solution and a great feature to add to the JobsTable.

sdc50:
replaced by #837

PR #704, opened 2021-07-08
- Add support to Cesium ion asset.

htran84:
- Add ability to pass in models and primitives into base map_view so it can be used in cesium map view.

PR #707, opened 2021-07-15
Add support cesium ion primitives

htran84:


swainn:
@sdc50 this is a new version of #704 rebased to take care of the merge conflicts. The failing test is not related to this PR. I think we should address it in a separate PR.

swainn (approved):

tethys_gizmos/gizmo_options/cesium_map_view.py
```diff
@@ -461,8 +471,8 @@ class CesiumMapView(TethysGizmoOptions):
     # Set Cesium Default Release Version.
     cesium_version = "1.63.1"
 
-    def __init__(self, options={}, globe={}, view={}, layers=[], entities=[], terrain={}, models={}, clock={},
-                 height='100%', width='100%', draw=False, attributes={}, classes='', cesium_ion_token=''):
+    def __init__(self, options={}, globe={}, view={}, layers=[], entities=[], primitives=[], terrain={}, models={},
```
sdc50, line 474: There could potentially be some problems with the empty dicts/lists in the function definition. It is better practice to do:

```python
def __init__(self, options=None ...):
    self.options = options or {}
```

That way each instance gets its own default empty dict/list rather than having all instances share the same one.

sdc50 (changes requested):

swainn (approved):

PR #711, opened 2021-07-29
Updates ol CDN URL and fixes 1 compatibility issue

shawncrawley:
There are very likely other compatibility issues that would need to be addressed if we truly wanted Tethys Core robust enough to allow for ANY version of OpenLayers. But I only quickly came upon and addressed the one issue with "ol.inherits".

### DIFF
--- a/tethys_gizmos/gizmo_options/map_view.py
+++ b/tethys_gizmos/gizmo_options/map_view.py
@@ -313,7 +313,7 @@ class MapView(TethysGizmoOptions):
     """  # noqa: E501
     gizmo_name = "map_view"
     ol_version = '5.3.0'
-    cdn = 'https://cdn.jsdelivr.net/npm/openlayers@{version}/dist/ol{debug}.{ext}'
+    cdn = 'https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v{version}/{folder}/ol{debug}.{ext}'
     alternate_cdn = 'https://cdnjs.cloudflare.com/ajax/libs/openlayers/{version}/ol{debug}.{ext}'
     local_url = 'tethys_gizmos/vendor/openlayers/{version}/ol.{ext}'
 
@@ -355,6 +355,7 @@ class MapView(TethysGizmoOptions):
         openlayers_library = cls.static_url().format(
             version=cls.ol_version,
             debug=cls.debug(),
+            folder='build',
             ext='js'
         )
 
@@ -378,6 +379,7 @@ class MapView(TethysGizmoOptions):
         openlayers_css = cls.static_url().format(
             version=cls.ol_version,
             debug=cls.debug(),
+            folder='css',
             ext='css'
         )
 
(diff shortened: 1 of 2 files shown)


tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
```diff
@@ -6,6 +6,14 @@
...
  *****************************************************************************/
 
+ // Backward/forward compatability for OpenLayers
+if (typeof ol.inherits === 'undefined') {
+    ol.inherits = function (child, parent) {
+        child.prototype = Object.create(parent.prototype);
+        child.prototype.constructor = child;
+    }
+}
+
```
swainn, line 16: Can you elaborate on why this is needed? Does something break in the Tethys MapView in newer versions?

tethys_gizmos/gizmo_options/map_view.py
```diff
@@ -313,7 +313,7 @@ class MapView(TethysGizmoOptions):
     """  # noqa: E501
     gizmo_name = "map_view"
     ol_version = '5.3.0'
-    cdn = 'https://cdn.jsdelivr.net/npm/openlayers@{version}/dist/ol{debug}.{ext}'
+    cdn = 'https://cdn.jsdelivr.net/gh/openlayers/openlayers.github.io@master/en/v{version}/{folder}/ol{debug}.{ext}'
```
swainn, line 316: Are there other options you would pass in for `folder`? It seems to me that we can do without `folder` being a variable and just hard-code it to be "build" in the URL.

swainn (changes requested):
Take a look at my comments.

tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
shawncrawley: The "inherits" method had been deprecated and was removed in version OpenLayers version 6, but it is used in the Tethys MapView. See notes [here](https://github.com/openlayers/openlayers/blob/main/changelog/upgrade-notes.md#removal-of-deprecated-methods). I know that allowing users to input their own version will surely one day result in more and more issues with the Tethys MapView, but I figured we could try and cover the ones that are blatantly obvious at this point - which is only this one as far as I found.

tethys_gizmos/gizmo_options/map_view.py
shawncrawley: For JavaScript, this folder is "build", but for CSS it is "css". See lines below.

swainn: Ah, I see now. Sorry, I didn't look closely enough I suppose.

tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
swainn: Makes sense. Sounds like maybe at the ECMA/JavaScript level? As far as supporting future versions, I guess that's the risk the user takes using a version that is not the version we default to, provided we are doing a reasonable job at keeping up to date. Could you add a note to the documentation on this feature that indicates this? Sort of a "use at your own risk" or advise not to update more than minor versions type of thing?

swainn (changes requested):

swainn (approved):

PR #742, opened 2021-11-16
add docs for cesium reInitializeMap api

sdc50:


swainn (approved):

swainn:
@sdc50 this will be a great addition to the Cesium Map View! Thanks for the contribution.

PR #753, opened 2022-01-03
Django 3.2 Upgrade

swainn:
Upgrade Tethys Platform to be compatible with Django 3.2. Which included the following changes:

Django Upgrade

* Upgrading to Channels 3, which makes use of the built-in ASGI support that was introduced in Django 3. Both Django 3 and Channels 3 now implement the ASGI 3 standard, which introduced a few breaking changes.

Model and Admin Changes
* Changes to Model admin definitions (add obj argument to has_add_permission)
* Migrating to generic Django model JSONField instead of PostgreSQL-specific JSONField.
* Set the DEFAULT_AUTO_FIELD setting to BigAutoField and create migrations to migrate PKs accordingly.

Web Sockets and URLs
* Refactored Web Socket implementation to use new consumer syntax
* Web Socket endpoints are now namespaced with "apps" (e.g.: "/apps/my-first-app/home/ws").
* Refactored how URLs are generated for Web Sockets (twice) so that they are the same as normal URLs (needing "/apps/<app-root>/" prepended to them).
* HTTP endpoints generated by a websocket/bokeh UrlMaps are now loaded with other HTTP URL patterns for apps.
* Moved logic from `tethys_portal\routing.py` to `tethys_apps\urls.py` and `tethys_portal\asgi.py` to simplify and streamline where web socket URLs are generated.

Dependencies
* Updated `environment.yml` to reflect new dependencies for upgrade (Django 3.2, Channels 3, and Daphne 3)
* Removed constraints on dependencies that are no longer necessary (psycopg2 and bokeh)
* Prioritized conda-forge over tethysplatform to pull newly packaged Django dependencies from conda-forge (all Django dependencies added now).

Deprecations
* Removed the deprecated workspace methods from apps.
* Refactored JobsManager to use the internal _get_user_workspace method instead of the removed app method.

Documentation
* Updated documentation build dependencies to be consistent with Tethys 4 dependencies (Python, Django, etc.)
* Resolved build issues after upgrading dependencies
* Removed/updated references to the old workspace methods.

Tethys CI
* Refactored the install_tethys.sh script (which is still used by our CI) to use mamba to create the Tethys environment for a speed up in the Test and Conda Build jobs.
* Refactored CI some to fix the Conda build, which was failing after upgrading dependencies.

**IMPORTANT: The Django implementation of Bokeh is not being maintained and has not been upgraded to support Django 3 (see: [Update bokeh.server.django to work with Django 3.2 #11148](https://github.com/bokeh/bokeh/issues/11148)). The Bokeh handler implementation will not work in Tethys 4 until Bokeh upgrades to support Django 3. The Bokeh implementation remains in place in this PR, but is not functioning properly due to Bokeh's failure to maintain their Django support.**


swainn:
@sdc50 this should be ready for review. This is a functioning version of Tethys with Django 3 and other upgraded dependencies.

sdc50 (approved):

sdc50:
@swainn  Wow! This was a lot of work. Looks good to me. @shawncrawley Do you want to finish reviewing before we merge?

swainn:
@msouff I'd be interested on your review on this PR since you implemented the initial web socket capability in Tethys

PR #759, opened 2022-01-18
Scheduler App Settings

swainn:
Adds a new type of app setting, Scheduler Setting, that can be used to assign HTCondor and Dask schedulers to an app as other services are.

image

* The Scheduler Settings accept the existing CondorScheduler and DaskScheduler objects, which will work with the existing JobManager APIs.
* New `SchedulerSetting` objects that can be used to define scheduler settings. Currently support two types of engines: 'HTCondor' and 'Dask'.
* New `scheduler_settings` override method added to TethysAppBase class that expects a list of `SchedulerSetting` objects to be returned.
* New `get_scheduler` method added to the TethysAppBase class for convenient means of accessing schedulers
* Jobs, settings, and related documentation updated
* Updated / Added Tests

Other Changes:
* Adds "Port" field to the CondorScheduler so it can be specified
* Hides password and private_key_pass fields on the list/table view of condor schedulers in the admin pages.


swainn:
@sdc50 @shawncrawley @msouff This is ready for review. Tests are failing b/c until the new version of CondorPy becomes available on Conda Forge: https://github.com/conda-forge/staged-recipes/pull/17592

sdc50 (approved):

sdc50:
@swainn This should be named `Scheduler App Settings - and documentation/admin pages overhaul`! It looks good to me. I saw one stickler error, but other than that it's ready to go.

swainn:
> @swainn This should be named `Scheduler App Settings - and documentation/admin pages overhaul`! It looks good to me. I saw one stickler error, but other than that it's ready to go.

Haha, yeah. There were a lot of references to the old pattern of using schedulers, and then I noticed that we still had a lot of references to the old Tethys Compute functionality that we removed a while ago, so it turned into a pretty big overhaul.

swainn:
Had to pin openssl to <3.0.0 to get the MacOS tests to pass. See: #761

Issue #763, opened 2022-02-02
Spatial Dataset Service created w/ CLI does not work until manually saved in admin pages


swainn:
Tethys Version: 3.4.1

If the `tethys services create` command is used to create a spatial dataset service like so:

```
tethys services create spatial -t GeoServer -n tethys_geoserver -c admin:geoserver@http://localhost:8181/geoserver/rest/ -p https://www.example.com:8181/geoserver/rest/
```

attempting to retrieve the engine returns `None`, which can result in errors like the following:

```
2022-02-02 16:24:10,875 ERROR    Internal Server Error: /apps/sirba-forecast/sirba_forecast/
Traceback (most recent call last):
  File "/opt/conda/envs/tethys/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/opt/conda/envs/tethys/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/opt/conda/envs/tethys/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "./tethys_apps/decorators.py", line 34, in wrapper
    return controller(request, *args, **kwargs)
  File "/opt/conda/envs/tethys/lib/python3.7/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/opt/conda/envs/tethys/lib/python3.7/site-packages/tethysapp_sirba_forecast-0.0.1-py3.7.egg/tethysapp/sirba_forecast/controllers.py", line 197, in home
    response = geoserver_engine.list_workspaces()
AttributeError: 'NoneType' object has no attribute 'list_workspaces'
```

However, this can be fixed by opening the Spatial Dataset Service in the admin pages and saving it again. ``tethys services create`` seems to be missing some important step.

This issue can be reproduced using the following docker-compose: https://github.com/Aquaveo/tethys_portal_docker_sirba

swainn:
The `tethys service create` command is assigning the wrong value for engine here:

https://github.com/tethysplatform/tethys/blob/6266f2cb1360100894ffecd6fdbc5dffaa79ac32/tethys_cli/services_commands.py#L197

It needs to assign values as is done here: 

[...]

PR #764, opened 2022-02-02
Fix for #763

swainn:
See: #763


msouff (approved):
I see datasets services had the same issue. Good thing you caught it.

swainn:
> I see datasets services had the same issue. Good thing you caught it.

Indeed it did. Easy enough to fix though.

sdc50 (approved):

PR #766, opened 2022-02-02
fix URL routing for http handler URLS

sdc50:
Changes the way that `http_handler_patterns` are routed to allow the ASGI consumers to be called correctly.

### DIFF
--- a/tethys_apps/base/app_base.py
+++ b/tethys_apps/base/app_base.py
@@ -1085,7 +1085,7 @@ class TethysAppBase(TethysBase):
 
         Args:
           name(string): Name of the PersistentStoreConnectionSetting as defined in app.py.
-          as_re_path(bool): Return SQLAlchemy URL object instead of engine object if True. Defaults to False.
+          as_url(bool): Return SQLAlchemy URL object instead of engine object if True. Defaults to False.
           as_sessionmaker(bool): Returns SessionMaker class bound to the engine if True.  Defaults to False.
 
         Returns:
@@ -1125,7 +1125,7 @@ class TethysAppBase(TethysBase):
 
         Args:
           name(string): Name of the PersistentStoreConnectionSetting as defined in app.py.
-          as_re_path(bool): Return SQLAlchemy URL object instead of engine object if True. Defaults to False.
+          as_url(bool): Return SQLAlchemy URL object instead of engine object if True. Defaults to False.
           as_sessionmaker(bool): Returns SessionMaker class bound to the engine if True.  Defaults to False.
 
         Returns:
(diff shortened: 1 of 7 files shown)


sdc50:
Note that this PR is necessary, but not sufficient to fix the Bokeh integration. There are also changes on the Bokeh side that are needed. I've successfully made those changes locally, but haven't pushed a PR to Bokeh yet.

swainn:
> Note that this PR is necessary, but not sufficient to fix the Bokeh integration. There are also changes on the Bokeh side that are needed. I've successfully made those changes locally, but haven't pushed a PR to Bokeh yet.

Awesome! Hopefully, they'll merge the changes on their end!

tethys_apps/base/app_base.py
```diff
@@ -1085,7 +1085,7 @@ def get_persistent_store_connection(cls, name, as_url=False, as_sessionmaker=Fal
 
         Args:
           name(string): Name of the PersistentStoreConnectionSetting as defined in app.py.
-          as_url(bool): Return SQLAlchemy URL object instead of engine object if True. Defaults to False.
+          as_re_path(bool): Return SQLAlchemy URL object instead of engine object if True. Defaults to False.
```
swainn, line 1088: Opps. This change is not necessary I hope. :)

tethys_apps/base/app_base.py
```diff
@@ -1125,7 +1125,7 @@ def get_persistent_store_database(cls, name, as_url=False, as_sessionmaker=False
...
```
swainn, line 1128: Another instance of a docs string changed inadvertently.

tethys_apps/urls.py
```diff
@@ -8,46 +8,41 @@
...
-    url(r'^send-beta-feedback/$', send_beta_feedback_email, name='send_beta_feedback'),
+    re_path(r'^$', library, name='app_library'),
+    re_path(r'^send-beta-feedback/$', send_beta_feedback_email, name='send_beta_feedback'),
 ]
 
 # Append the app urls urlpatterns
 harvester = SingletonHarvester()
 normal_url_patterns = harvester.get_url_patterns()
 handler_url_patterns = harvester.get_handler_patterns()
 app_url_patterns = normal_url_patterns['app_url_patterns']
```
swainn, line 27: Is `app_url_patterns` used anywhere else other than the loop below? If not we should inline it in the loop as we do with `http_handler_patterns` for consistency.

swainn (changes requested):
See comments.

tethys_apps/base/app_base.py
sdc50: Good catch!

swainn (approved):

PR #767, opened 2022-02-02
Make Using CDN Optional

sdc50:
Changes the way that `http_handler_patterns` are routed to allow the ASGI consumers to be called correctly.


sdc50:
There are sill a few of pending issues/questions about how to implement this PR:

- gizmos that allow the versions of dependencies to be specified as a class variable. 
  -  This works fine when using a CDN, but if dependencies are installed with `npm` then there is only one version
- conda package/production installations
  - I currently have the npm packages downloaded into the `tethys_portal/static/` directory. This is working great in dev. I'm not sure how this will work with a conda package where that dir is in `site-packages/`. Also, I'm assuming that `collectstatic` will work since it's in the `tethys_portal/static/` dir.
- esri_map
  -  the version of the js api that we are using is so old that it is not available on npm (in a usable format). I think I'll need to update this to the latest version
-  bokeh
  - This is using the bokeh CDN setting in the Gizmo. I'm planning to change this to use a setting from `settings.py`
- D3
  - this is also using an old version. The `jobs_table` gizmo was using a different version than the plots gizmo, but it seems to work fine with the older version.

swainn:
@sdc50 is this ready for an initial review?

tethys_portal/dependencies.py
```diff
@@ -0,0 +1,245 @@
...
+    def default_debug_path_converter(path):
+        return path.replace('min.', '')
+
+    def get_custom_version_url(self, url_type, version, debug=None):
+        if version == self.version:
+            pass
+        elif not self.use_cdn:
+            logger.warning(
+                f'The "STATICFILES_USE_CDN" setting is set to False, so the custom version "{self.npm_name}={version}" '
+                f'is not supported. Using the default version ({self.version}) instead.'
```
swainn, line 55: Why are custom versions not supported for npm?

docs/tethys_portal/configuration.rst
```diff
@@ -57,6 +57,7 @@ The following is a list of keys that can be added to the :file:`portal_config.ym
     * **ENABLE_RESTRICTED_APP_ACCESS**: app access can be restricted based on user object permissions when ``True``. Defaults to ``False``. If ``ENABLE_OPEN_PORTAL`` is set to ``True`` this setting has no effect. That is, users will have unrestricted access to apps independently of the value of this setting.
     * **TETHYS_WORKSPACES_ROOT**: location to which app workspaces will be synced when ``tethys manage collectworkspaces`` is executed. Gathering all workspaces to one location is recommended for production deployments to allow for easier updating and backing up of app data. Defaults to :file:`<TETHYS_HOME>/workspaces`.
     * **STATIC_ROOT**: the Django `STATIC_ROOT <https://docs.djangoproject.com/en/2.2/ref/settings/#static-root>`_ setting. Defaults to :file:`<TETHYS_HOME>/static`.
+    * **STATICFILES_USE_CDN**: uses a content delivery network (CDN) to serve JavaScript dependencies when ``True``. Defaults to ``True``. When set to ``False`` then you must run :cmd:`tethys gen vendor_js` to download the JS dependencies locally to be served by Tethys.
```
swainn, line 60: It would make more sense to me if this was renamed so that you are “enabling” installing with NPM rather than disabling the default behavior? Perhaps something like `STATICFILES_USE_NPM = True`.

swainn (changes requested):
Very excited for this. See comments.

tethys_portal/dependencies.py
sdc50: Only one version of the npm package is downloaded (the versions specified in the `vendor_static_dependencies` dict below. If we wanted to support custom versions of something then the user would have to manually download that version of the dependency. It may work, but would probably require some testing.

tethys_cli/gen_commands.py
```diff
@@ -21,6 +24,7 @@
 
 import tethys_portal
 from tethys_apps.utilities import get_tethys_home_dir, get_tethys_src_dir
+from tethys_portal.dependencies import dependencies as vendor_static_dependencies
```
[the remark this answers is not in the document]
sdc50: maybe not rename the module, but at least the dictionary inside the module that's being imported here.

tethys_portal/dependencies.py
sdc50: I changed the behavior so that if a custom version is requested  when the `USE_NPM` option is `True` then a warning is issued saying it will attempt to use the CDN.

docs/tethys_portal/configuration.rst
sdc50: Yes, using the CDN is the default. I've made the name change that you suggested, and toggled all of the boolean values.

sdc50:
@swainn I addressed you comments, and removed the remaining `vendor/` files. I think this is ready to go now.

swainn (approved):
Looks good to me.

swainn:
@sdc50 it feels good to be dropping all those vendored packages. :D

PR #775, opened 2022-02-19
Controller decorator

### LINKED ISSUE #779: Changing the `root_url` attribute of an app breaks the index URL
If I wanted to change the `root_url` from the value that the scaffold generates it would break the home view of the app. 

For example, if I change the `BokehTutorial` app's `root_url` to `'bokeh-demo'` :

```python
class BokehTutorial(TethysAppBase):
    """
    Tethys app class for Bokeh Tutorial.
    """

    name = 'Bokeh Tutorial'
    index = 'bokeh_tutorial:home'
    root_url = 'bokeh-demo'
```
I would get this error:

image

This is because  Tethys uses the `root_url` attribute to derive the URL namespace that is actually used to register the URLs. So for this to work we would need to also change the `index` attribute to:

```python
index = 'bokeh_demo:home`
```

This doesn't seem intuitive, so I propose that we deprecate defining the `index` attribute with the URL namespace and have Tethys add it in with the namespace that it is actually going to use to register the URLs.

As part of deprecating the current way the `index` attribute is defined we could provide backwards compatibility for one release cycle and print a deprecation warning similar to this:

image

sdc50:
Fixes #779


sdc50:
I've got a pretty good start to this. There are a few design decisions that need to be hashed out. I've written some basic docs in the doc string that hopefully explain how the new ``controller`` decorator can be used:

image
image
image
image

sdc50:
I had to move several decorators that were in different `tethys_*` packages into `tethys_apps` to avoid import errors. Currently the ``enforce_quotas`` argument is not working due to circular imports. To get it work I think I would need to pull the ``decorator`` module out of `tethys_apps` into its own package, or I would need to make `tethys_quotas` a sub-packaged of `tethys_apps`.

tethys_apps/base/controller.py
```diff
@@ -18,3 +48,286 @@ def as_controller(cls, **kwargs):
...
+
+        # additional kwargs to pass to TethysController.as_controller
+        **kwargs
+) -> Callable:
+    """
+    Decorator to register a function or TethysController class as a controller
+    (by automatically registering a UrlMap for it).
+
+    Args:
+        name: Name of the url map. Letters and underscores only (_). Must be unique within the app. The default is the name of the function being decorated.
```
swainn, line 94: What if you have two functions with the same name in different submodules of controllers?

* controllers.resource.view
* controllers.workflows.view

tethys_apps/base/controller.py
```diff
@@ -18,3 +48,286 @@ def as_controller(cls, **kwargs):
...
+        url: URL pattern to map the endpoint for the controller or consumer. If a `list` then a seperate UrlMap is generated for each URL in the list. The first URL is given `name` and subsequent URLS are named `name`_1, `name`_2 ... `name`_n. Can also be passed as dict mapping names to URL patterns. In this case the `name` argument is ignored.
+        protocol: 'http' for consumers or 'websocket' for consumers. Default is http.
+        regex: Custom regex pattern(s) for url variables. If a string is provided, it will be applied to all variables. If a list or tuple is provided, they will be applied in variable order.
+        handler: Dot-notation path a handler function. A handler is associated to a specific controller and contains the main logic for creating and establishing a communication between the client and the server.
```
swainn, line 98: handler decorator? :)

tethys_apps/base/controller.py
```diff
@@ -18,3 +48,286 @@ def as_controller(cls, **kwargs):
...
+        **kwargs
+) -> Callable:
+    """
+    Decorator to register a function or TethysController class as a controller
+    (by automatically registering a UrlMap for it).
+
+    Args:
+        name: Name of the url map. Letters and underscores only (_). Must be unique within the app. The default is the name of the function being decorated.
+        url: URL pattern to map the endpoint for the controller or consumer. If a `list` then a seperate UrlMap is generated for each URL in the list. The first URL is given `name` and subsequent URLS are named `name`_1, `name`_2 ... `name`_n. Can also be passed as dict mapping names to URL patterns. In this case the `name` argument is ignored.
+        protocol: 'http' for consumers or 'websocket' for consumers. Default is http.
```
swainn, line 96: Have you tested this with websockets or bokeh handlers?

tethys_apps/base/controller.py
```diff
@@ -18,3 +48,286 @@ def as_controller(cls, **kwargs):
...
+            # Note that if both the ``app_workspace`` and ``user_workspace`` arguments are passed to the controller
+            # decorator, then "app_workspace" should preceed "user_workspace" in the function argument list,
+            # and both should be directly after the "request" argument.
+            ...
+
+        @controller(
+            login_required=False,
+        )
+        def my_app_controller(request):
+            # Note that ``login_required`` is ``True`` by default.
```
swainn, line 195: "# Note that ``login_required`` is ``True`` by default (recommended). However, the login requirement is automatically dropped on all controllers when ``OPEN_PORTAL_MODE`` is enabled on the portal."

tethys_sdk/services.py
```diff
@@ -1 +1 @@
...
# DO NOT ERASE
from tethys_services.utilities import (
    list_dataset_engines,
    get_dataset_engine,
    list_spatial_dataset_engines,
    get_spatial_dataset_engine,
    list_wps_service_engines,
    get_wps_service_engine,
    ensure_oauth2,
)
```
swainn, line 1: Not sure what is going on here.

swainn (changes requested):
In addition to passing tests and linting, see comments. I'm very much looking forward to this feature.

tethys_apps/base/controller.py
sdc50: You would have to manually provide a name for at least one of them.

tethys_apps/base/controller.py
sdc50: Wait, did I mis-type something, or are you volunteering to create a handler decorator?

tethys_apps/base/controller.py
sdc50: Also, I pulled this docstring from the `UrlMapBase` class, but I think there is a typo. Should it be:

```suggestion
        protocol: 'http' for controllers or 'websocket' for consumers. Default is http.
```

tethys_sdk/services.py
sdc50: I'm not sure either. Is it possible this file was created on Windows and had `\n\r` characters?

tethys_apps/base/controller.py
sdc50: I'm now catching that case and providing a different name, but printing out a big warning message:

image

tests/apps/tethysapp-test_app/tethysapp/test_app/controllers.py
```diff
@@ -109,6 +112,11 @@ def callback(attr: str, old: Any, new: Any) -> None:
     doc.add_root(column(slider, plot))
 
 
+@controller(
```
swainn, line 115: Change to the new `consumer` decorator.

swainn (changes requested):
Added a few more comments.

Also, implement what we discussed on Friday.

sdc50:
@swainn This is ready for final review. I ended up just creating a ``handler`` decorator. 

It seems like the conda build is failing due to memory issues. I'm not sure how to go about addressing that with GitHub Actions.

tethys_apps/base/controller.py
```diff
@@ -18,3 +43,529 @@ def as_controller(cls, **kwargs):
...
+
+        with_request: bool = False,
+        with_workspaces: bool = False,
+        **kwargs,
+):
+    """
+    Decorator to register a handler function and connect it with a controller function
+    (by automatically registering a UrlMap for it).
+
+    Args:
```
swainn, line 390: Need docstrings for `with_request` and `with_workspaces` arguments.

tethys_apps/base/controller.py
```diff
@@ -18,3 +43,529 @@ def as_controller(cls, **kwargs):
...
+        )
+        def my_app_handler(document):
+            ...
+
+        ------------
+
+        @handler(
+            with_request=True
+        )
+        def my_app_handler(document):
```
swainn, line 452: `document` -> `doc` or vica versa down below.

tethys_apps/base/controller.py
```diff
@@ -18,3 +43,529 @@ def as_controller(cls, **kwargs):
...
+            # attribute available when using "with_request" argument
+            request = doc.request
+            ...
+
+        ------------
+
+        @handler(
+            with_workspaces=True
+        )
+        def my_app_handler(document):
```
swainn, line 462: `document` -> `doc` or vica versa down below.

tethys_cli/scaffold_templates/app_templates/default/setup.py_tmpl
```diff
@@ -1,17 +1,16 @@
 from setuptools import setup, find_namespace_packages
-from tethys_apps.app_installation import find_resource_files
+from tethys_apps.app_installation import find_all_resource_files
+from tethys_apps.base.app_base import TethysAppBase
 
 # -- Apps Definition -- #
 app_package = '{{project}}'
-release_package = 'tethysapp-' + app_package
+release_package = f'TethysAppBase.package_namespace-{app_package}'
```
swainn, line 7: Need braces around `TethysAppbase.package_namespace`: `f'{TethysAppBase.package_namespace}-{app_package}`

swainn (changes requested):
This is great @sdc50! The new decorators will be an excellent addition. Just a few minor changes to make and this will be done!

tethys_apps/base/app_base.py
```diff
@@ -1537,3 +1563,109 @@ def post_delete_app_workspace(cls):
...
+            nm = importlib.import_module(f'.{sm.name}', m.__name__)
+            modules.append(nm)
+            if sm.ispkg:
+                modules.extend(get_all_submodules(nm))
+    except AttributeError:
+        pass
+    return modules
+
+
+def write_warning(msg):
```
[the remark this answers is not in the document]
sdc50: I can't import it into this module because it would create a circular import. I could import it into each function that uses it.

tethys_apps/base/app_base.py
```diff
@@ -1537,3 +1563,109 @@ def post_delete_app_workspace(cls):
...
+    FG_YELLOW = '\033[33m'
+    ALL_OFF = '\033[0m'
+    print(f'{FG_YELLOW}{msg}{ALL_OFF}')
+
+
+def register_controllers(root_url: str, modules: Union[str, list, tuple], index: str = None) -> list:
```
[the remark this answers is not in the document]
sdc50: again it is in here to prevent circular imports. But again I could try importing it in the method that uses it.

tethys_apps/base/app_base.py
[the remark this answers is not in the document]
swainn: I think I'd prefer that over maintaining two duplicate functions.

tethys_apps/base/app_base.py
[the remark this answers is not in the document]
swainn: Dang those circular imports. This one is trickier b/c users will potentially be using it. I suspect if we move it then they will also need to deal with circular imports. If that is the case, then let's leave it.

swainn:
> @swainn This is ready for final review. I ended up just creating a `handler` decorator.
> 
> It seems like the conda build is failing due to memory issues. I'm not sure how to go about addressing that with GitHub Actions.

Re: conda build. I've noticed this and I'm not sure how to handle it either. Perhaps we could try building with mamba's builder? Let's not worry about it for this PR though. I'd like to get it merged sooner than later..

tethys_apps/base/app_base.py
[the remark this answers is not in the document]
sdc50: If uses ever need to use this function then they would import it from `tethys_sdk.routing` so I don't think they would run into the circular import issues.

tests/unit_tests/test_tethys_apps/test_base/test_app_base.py
```diff
@@ -5,6 +5,10 @@
 from django.db.utils import ProgrammingError
 from django.test import RequestFactory
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
+
+import tethys_apps.base.app_base
+import tethys_apps.base.controller
+import tethys_apps.utilities
```
sdc50, line 11:
```suggestion
```

tethys_apps/base/app_base.py
```diff
@@ -26,44 +26,72 @@
 
 from bokeh.server.django.consumers import AutoloadJsConsumer
 
+# import for type hinting
```
sdc50, line 29:
```suggestion
```

tethys_apps/base/app_base.py
```diff
@@ -26,44 +26,72 @@
...
+
```
sdc50, line 30:
```suggestion
```

tests/unit_tests/test_tethys_apps/test_base/test_app_base.py
```diff
@@ -5,6 +5,9 @@
 from django.db.utils import ProgrammingError
 from django.test import RequestFactory
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
+
+import tethys_apps.base.app_base
+import tethys_apps.base.controller
```
sdc50, line 10:
```suggestion
```

swainn (approved):

PR #781, opened 2022-03-21
Tethys 4 manage command

msouff:
- added tethys manage path to print manage.py path
- added support to call django manage commands through tethys manage
- added support for unknown args to tethys commands

### DIFF
--- a/tethys_cli/cli_helpers.py
+++ b/tethys_cli/cli_helpers.py
@@ -52,11 +52,8 @@ def run_process(process):
         set_testing_environment(False)
 
 
-def load_apps(silent=False):
-    if silent:
-        stdout = sys.stdout
-        sys.stdout = open(os.devnull, 'w')
-        django.setup()
-        sys.stdout = stdout
-    else:
-        django.setup()
+def load_apps():
+    stdout = sys.stdout
+    sys.stdout = open(os.devnull, 'w')
+    django.setup()
+    sys.stdout = stdout
--- a/tethys_cli/cli_helpers.py
+++ b/tethys_cli/cli_helpers.py
@@ -1,8 +1,8 @@
 import os
+import sys
 import subprocess
 
 import django
-from django.core.management import get_commands
 
 from tethys_apps.base.testing.environment import set_testing_environment
 from tethys_apps.utilities import get_tethys_src_dir
@@ -40,12 +40,6 @@ def get_manage_path(args):
     return manage_path
 
 
-def get_django_manage_commands():
-    manage_commands = sorted(list(get_commands().keys()))
-
-    return manage_commands
-
-
 def run_process(process):
     # Call the process with a little trick to ignore the keyboard interrupt error when it happens
     try:
@@ -58,5 +52,11 @@ def run_process(process):
         set_testing_environment(False)
 
 
-def load_apps():
-    django.setup()
+def load_apps(silent=False):
+    if silent:
+        stdout = sys.stdout
+        sys.stdout = open(os.devnull, 'w')
+        django.setup()
+        sys.stdout = stdout
+    else:
+        django.setup()
--- a/tethys_cli/cli_helpers.py
+++ b/tethys_cli/cli_helpers.py
@@ -38,11 +38,13 @@ def get_manage_path(args):
 
     return manage_path
 
+
 def get_django_manage_commands():
     from django.core.management import get_commands
     manage_commands = sorted(list(get_commands().keys()))
-    
+
     return manage_commands
+
 
 def run_process(process):
     # Call the process with a little trick to ignore the keyboard interrupt error when it happens
--- a/tethys_cli/cli_helpers.py
+++ b/tethys_cli/cli_helpers.py
@@ -2,7 +2,6 @@ import os
 import subprocess
 
 import django
-from django.core.management import ManagementUtility
 
 from tethys_apps.base.testing.environment import set_testing_environment
 from tethys_apps.utilities import get_tethys_src_dir
@@ -39,12 +38,11 @@ def get_manage_path(args):
 
     return manage_path
 
-def get_manage_commands():
-    manage_utils = ManagementUtility()
-    manage_commands = manage_utils.main_help_text(commands_only=True).split('\n')
+def get_django_manage_commands():
+    from django.core.management import get_commands
+    manage_commands = sorted(list(get_commands().keys()))
     
     return manage_commands
-    
 
 def run_process(process):
     # Call the process with a little trick to ignore the keyboard interrupt error when it happens
--- a/tethys_cli/cli_helpers.py
+++ b/tethys_cli/cli_helpers.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 
 import django
+from django.core.management import get_commands
 
 from tethys_apps.base.testing.environment import set_testing_environment
 from tethys_apps.utilities import get_tethys_src_dir
@@ -40,7 +41,6 @@ def get_manage_path(args):
 
 
 def get_django_manage_commands():
-    from django.core.management import get_commands
     manage_commands = sorted(list(get_commands().keys()))
 
     return manage_commands
--- a/tethys_cli/cli_helpers.py
+++ b/tethys_cli/cli_helpers.py
@@ -2,6 +2,7 @@ import os
 import subprocess
 
 import django
+from django.core.management import ManagementUtility
 
 from tethys_apps.base.testing.environment import set_testing_environment
 from tethys_apps.utilities import get_tethys_src_dir
@@ -38,6 +39,12 @@ def get_manage_path(args):
 
     return manage_path
 
+def get_manage_commands():
+    manage_utils = ManagementUtility()
+    manage_commands = manage_utils.main_help_text(commands_only=True).split('\n')
+    
+    return manage_commands
+    
 
 def run_process(process):
     # Call the process with a little trick to ignore the keyboard interrupt error when it happens
--- a/tethys_cli/db_commands.py
+++ b/tethys_cli/db_commands.py
@@ -380,6 +380,6 @@ DB_COMMANDS = dict(
 )
 
 
-def db_command(args, unknown_args=""):
+def db_command(args, unknown_args=None):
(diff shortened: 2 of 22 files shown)


tethys_cli/cli_helpers.py
```diff
@@ -39,6 +40,12 @@ def get_manage_path(args):
     return manage_path
 
 
+def get_django_manage_commands():
+    manage_commands = sorted(list(get_commands().keys()))
+
+    return manage_commands
+
+
```
swainn, line 48: This function isn't generic enough to be here. Move it to the `manage_commands.py`

tethys_cli/manage_commands.py
```diff
@@ -8,22 +8,25 @@
...
 
 MANAGE_START = 'start'
 MANAGE_COLLECTSTATIC = 'collectstatic'
 MANAGE_COLLECTWORKSPACES = 'collectworkspaces'
 MANAGE_COLLECT = 'collectall'
 MANAGE_CREATESUPERUSER = 'createsuperuser'
-MANAGE_SHELL = 'shell'
+MANAGE_GET_PATH = 'path'
+
+TETHYS_COMMANDS = [MANAGE_START, MANAGE_COLLECTSTATIC, MANAGE_COLLECTWORKSPACES, MANAGE_COLLECT,
```
swainn, line 20: Have you tested collectstatic? Seems there would be a collision with the django `manage.py collecstatic` command.

tethys_cli/manage_commands.py
```diff
@@ -35,10 +38,13 @@ def add_manage_parser(subparsers):
     manage_parser.add_argument('-l', '--link', required=False, action='store_true',
                                help='Only used with collectstatic command. Link static directory to STATIC_ROOT '
                                     'instead of copying it. Not recommended.')
+    manage_parser.add_argument('--django-help', action='store_true',
+                               help='Display the help for specific scommands coming from Django. '
```
swainn, line 42: "scommands" -> "commands"

tethys_cli/db_commands.py
```diff
@@ -380,6 +380,6 @@ def process_args(args):
 )
 
 
-def db_command(args):
+def db_command(args, unknown_args=""):
```
swainn, line 383: Is `unknown_args` a string when it is given? If not, then you should set the default to `None` in all cases.

tethys_cli/manage_commands.py
```diff
@@ -35,10 +38,13 @@ def add_manage_parser(subparsers):
     manage_parser.add_argument('-l', '--link', required=False, action='store_true',
                                help='Only used with collectstatic command. Link static directory to STATIC_ROOT '
                                     'instead of copying it. Not recommended.')
+    manage_parser.add_argument('--django-help', action='store_true',
```
swainn, line 41: Is this necessary? Wouldn't `--help` just get passed through to the django command if they provided it? e.g.: `tethys manage shell --help` instead of `tethys manage shell --django-help`.

tethys_cli/manage_commands.py
```diff
@@ -96,5 +102,14 @@ def manage_command(args):
     elif args.command == MANAGE_CREATESUPERUSER:
         primary_process = ['python', manage_path, 'createsuperuser']
 
+    elif args.command == MANAGE_GET_PATH:
+        primary_process = ['python', '-c', f'print("{manage_path}")']
+
+    elif args.command in DJANGO_COMMANDS:
+        if args.django_help:
+            primary_process = ['python', manage_path, args.command, '--help']
```
swainn, line 110: Again, is this necessary?

tethys_cli/manage_commands.py
```diff
@@ -35,10 +38,13 @@ def add_manage_parser(subparsers):
...
                                help='Only used with collectstatic command. Link static directory to STATIC_ROOT '
                                     'instead of copying it. Not recommended.')
+    manage_parser.add_argument('--django-help', action='store_true',
+                               help='Display the help for specific scommands coming from Django. '
+                               'E.g. "tethys manage <django command> --django-help"')
     manage_parser.set_defaults(func=manage_command)
 
 
-def manage_command(args):
+def manage_command(args, django_args=""):
```
swainn, line 47: "django_args" -> "unknown_args" for consistency.

tethys_cli/gen_commands.py
```diff
@@ -333,7 +333,7 @@ def write_path_to_console(file_path):
 }
 
 
-def generate_command(args):
+def generate_command(args, unknow_args=""):
```
swainn, line 336: What happens when you provide an unknown argument now? Seems like it wouldn't catch it in the commands that don't use it. Perhaps good default behavior would be to combine these where `unknown_args` isn't used:

```
def generate_command(args, unknown_args=""):
    args = args + unknown_args
    ...
```

swainn: Also, "unknow_args" -> "unknown_args"

swainn (changes requested):
Take a look at the comments.

tethys_cli/cli_helpers.py
msouff: removed.

tethys_cli/manage_commands.py
msouff: the unknown commands are evaluated last.

tethys_cli/manage_commands.py
msouff: This is what I was trying too explain yesterday. "--help" calls the tethys help so I had to create the "--django-help" to avoid conflict and still be able to pull the help from django using tethys manage.

tethys_cli/db_commands.py
msouff: It is a list actually. I've updated the code to None.

tethys_cli/manage_commands.py
msouff: This is what I was trying too explain yesterday. "--help" calls the tethys help so I had to create the "--django-help" to avoid conflict and still be able to pull the help from django using tethys manage.

tethys_cli/manage_commands.py
msouff: done

tethys_cli/docker_commands.py
```diff
@@ -769,7 +769,7 @@ def docker_ip(containers=None):
         write_pretty_output(msg.format(name=container_metadata.display_name))
 
 
-def docker_command(args):
+def docker_command(args, docker_args=""):
```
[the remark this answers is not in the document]
msouff: done

tethys_cli/gen_commands.py
msouff: done

tethys_cli/scaffold_commands.py
```diff
@@ -116,7 +116,7 @@ def render_path(path, context):
     return path
 
 
-def scaffold_command(args):
+def scaffold_command(args, unknown_args=None):
```
swainn, line 119: What happens when you provide an unknown argument now? Seems like it wouldn't catch it in the commands that don't use it. Perhaps good default behavior would be to combine these where `unknown_args` isn't used:

```
def generate_command(args, unknown_args=""):
    if unknown_args is not None:
        args = args + unknown_args
    ...
```

swainn (changes requested):
One more comment

tethys_cli/manage_commands.py
```diff
@@ -35,10 +41,13 @@ def add_manage_parser(subparsers):
...
                                     'instead of copying it. Not recommended.')
-    manage_parser.set_defaults(func=manage_command)
+    manage_parser.add_argument('--django-help', action='store_true',
+                               help='Display the help for specific commands coming from Django. '
+                               'E.g. "tethys manage <django command> --django-help"')
+    manage_parser.set_defaults(func=manage_command, parsing_method='accepts_unknown_args')
 
 
-def manage_command(args):
+def manage_command(args, unknown_args=[]):
```
swainn, line 50: Better to default to None, then check for None and assign as list in the function like so:
```
def manage_command(args, unknown_args=None):
    if not unknown_args:
        unknown_args = []
```

tethys_cli/manage_commands.py
```diff
@@ -8,22 +8,28 @@
...
 MANAGE_CREATESUPERUSER = 'createsuperuser'
-MANAGE_SHELL = 'shell'
+MANAGE_GET_PATH = 'path'
 
 
 def add_manage_parser(subparsers):
-    # Setup start server command
+    # sub-command choices
+    TETHYS_COMMANDS = [MANAGE_START, MANAGE_COLLECTSTATIC, MANAGE_COLLECTWORKSPACES, MANAGE_COLLECT,
+                       MANAGE_CREATESUPERUSER, MANAGE_GET_PATH]
```
swainn, line 26: I realized the `MANAGE_CREATESUPERUSER` is just calling the `manage.py createsuperuser` command, so we should take the option out of `TETHYS_COMMANDS`.

tethys_cli/manage_commands.py
```diff
@@ -96,5 +105,14 @@ def manage_command(args):
     elif args.command == MANAGE_CREATESUPERUSER:
         primary_process = ['python', manage_path, 'createsuperuser']
 
+    elif args.command == MANAGE_GET_PATH:
+        primary_process = ['python', '-c', f'print("{manage_path}")']
```
swainn, line 109: Seems like overkill to spin up a subprocess to print the `manage_path`. You could just do this:

```
print(manage_path)
```

If you don't set `primary_process` it won't run it.

swainn (changes requested):
Last comments

tethys_cli/manage_commands.py
msouff: done

tethys_cli/manage_commands.py
msouff: done

tethys_cli/manage_commands.py
msouff: I got a kick out of this one too.

swainn (approved):

Issue #789, opened 2022-04-18
No Home Controller Found for Newly Scaffolded Apps in Tethys 4


swainn:
I installed a fresh Tethys 4 installation on Windows, scaffolded 3 apps, and installed them and I'm getting these warnings when Tethys starts:

![image](https://user-images.githubusercontent.com/5123221/163889014-e3f94db2-fcbb-41d4-a666-4d8bd7969891.png)

Visiting the app library page gives me this error:

<details>
```
ERROR:django.request:Internal Server Error: /apps/
Traceback (most recent call last):
  File "C:\Miniconda3\envs\tethys4\lib\site-packages\django\urls\base.py", line 71, in reverse
    extra, resolver = resolver.namespace_dict[ns]
KeyError: 'my_first_app'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Miniconda3\envs\tethys4\lib\site-packages\asgiref\sync.py", line 451, in thread_handler
    raise exc_info[1]
  File "C:\Miniconda3\envs\tethys4\lib\site-packages\django\core\handlers\exception.py", line 38, in inner
    response = await get_response(request)
  File "C:\Miniconda3\envs\tethys4\lib\site-packages\django\core\handlers\base.py", line 233, in _get_response_async
    response = await wrapped_callback(request, *callback_args, **callback_kwargs)
  File "C:\Miniconda3\envs\tethys4\lib\site-packages\asgiref\sync.py", line 414, in __call__
    ret = await asyncio.wait_for(future, timeout=None)
  File "C:\Miniconda3\envs\tethys4\lib\asyncio\tasks.py", line 408, in wait_for
    return await fut
  File "C:\Miniconda3\envs\tethys4\lib\site-packages\asgiref\current_thread_executor.py", line 22, in run
    result = self.fn(*self.args, **self.kwargs)
  File "C:\Miniconda3\envs\tethys4\lib\site-packages\asgiref\sync.py", line 455, in thread_handler
    return func(*args, **kwargs)
  File "c:\users\nswain\codes\tethysplatform\tethys_apps\decorators.py", line 36, in wrapper
    return controller(request, *args, **kwargs)
  File "C:\Miniconda3\envs\tethys4\lib\site-packages\django\contrib\auth\decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "c:\users\nswain\codes\tethysplatform\tethys_apps\views.py", line 75, in library
[...]

PR #790, opened 2022-04-19
Fixes Bad Import in App Scaffold

swainn:
This PR addresses issue #789 as follows:

* Fixes the import for the controller decorator in the app scaffold template
* Prints a warning and traceback for import errors that are not ModuleNotFoundErrors encountered while registering controllers:

image


swainn:
@sdc50 this is good to go. Docker build failing b/c patch is in the Map Layout PR

sdc50 (approved):
Assuming tests pass..

Issue #760, opened 2022-01-28
Logging configuration is broken


sdc50:
The Tethys `settings.py` file sets the `LOGGING_CONFIG` variable to `{}` by default. This setting should point to a callable that configures the loggers (https://docs.djangoproject.com/en/4.0/ref/settings/#logging-config).

sdc50:
As a temporary workaround you can add the following to your `portal_config.yml`

```yaml
LOGGING_CONFIG: logging.config.dictConfig
```

PR #792, opened 2022-04-19
Set appropriate default value for logging configuration

swainn:
Addresses issue #760

### DIFF
--- a/tethys_portal/settings.py
+++ b/tethys_portal/settings.py
@@ -134,7 +134,7 @@ LOGGERS.setdefault('tethys', {
     'handlers': ['console_verbose'],
     'level': 'INFO',
 })
-LOGGERS.setdefault('tethysapps', {
+LOGGERS.setdefault('tethysapp', {
     'handlers': ['console_verbose'],
     'level': 'INFO',
 })
--- a/tethys_portal/settings.py
+++ b/tethys_portal/settings.py
@@ -14,10 +14,10 @@ Settings for Tethys Platform
 This file contains default Django and other settings for the Tethys Platform.
 
 For more information on this file, see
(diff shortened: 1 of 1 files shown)


tethys_portal/settings.py
```diff
@@ -104,8 +101,8 @@
 DEFAULT_DB.setdefault('HOST', 'localhost')
 DEFAULT_DB.setdefault('PORT', 5436)
 
-LOGGING_CONFIG = portal_config_settings.pop('LOGGING_CONFIG', {})
-# See https://docs.djangoproject.com/en/1.8/topics/logging/#configuring-logging for more logging configuration options.
+# See https://docs.djangoproject.com/en/3.2/ref/settings/#logging-config for more logging configuration options.
+LOGGING_CONFIG = portal_config_settings.pop('LOGGING_CONFIG', 'logging.config.dictConfig')
```
sdc50, line 105: I think we should just remove this completely.

tethys_portal/settings.py
```diff
@@ -137,7 +134,7 @@
     'handlers': ['console_verbose'],
     'level': 'INFO',
 })
-LOGGERS.setdefault('tethys.apps', {
+LOGGERS.setdefault('tethysapps', {
```
sdc50, line 137:
```suggestion
LOGGERS.setdefault('tethysapp', {
```

this needs to be the same name as the namespace so that apps can just do:

```python
logger = logging.getLogger(__name__)
```

sdc50 (changes requested):

tethys_portal/settings.py
swainn: Done

sdc50 (approved):

Issue #685, opened 2021-04-16
Migrate away from setup.py conform with PEP 517


swainn:
We may need to migrate away from using setup.py as setuptools is no longer the official distribution mechanism.

See:

https://setuptools.readthedocs.io/en/latest/build_meta.html
https://www.python.org/dev/peps/pep-0517/

swainn:
PBR requires the use of the setup.py currently: https://docs.openstack.org/pbr/latest/user/features.html#setup-cfg

PR #794, opened 2022-04-21
Migrate to more standardized build metadata and mechanism

swainn:
Addresses Issue #685:

- Still using setuptools as build engine
- Migrate all metadata and configuration from setup.cfg to pyproject.toml
- Use setuptools_scm instead of pbr (pbr requires setup.py still)
- Removed dependence on setup.py, but setup.cfg still required for editable installs for now (see https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)
- Specify setuptools specific options in setup.cfg for now

Note: setuptools_scm does not support pre_versioning, which means the version of the current master branch changes from something like 4.0.0.dev17 to  3.4.1.dev17+g75925919.d20220421

Also Note: this PR does not address app installation with setup.py


swainn:
@sdc50 this should be good to go, though it would be good to rebase this one with the docker changes in the Map Layout PR after it is merged.

conda.recipe/meta.yaml
```diff
@@ -1,11 +1,11 @@
 # Conda Recipe for Tethys Platform
 # WARNING: THIS IS A GENERATED FILE. DO NOT EDIT.
 # TO CHANGE THIS FILE, SEE $TETHYS_SRC/tethys_cli/gen_templates/meta yaml
-{% set data = load_setup_py_data() %}
+
 
 package:
   name: tethys-platform
-  version: {{ data['version'] }}
+  version: 3.4.1.dev24+gb092a0af.d20220421
```
sdc50, line 8: Should this version be hard coded?

tethys_portal/_version.py
```diff
@@ -0,0 +1,5 @@
+# coding: utf-8
+# file generated by setuptools_scm
+# don't change, don't track in version control
```
sdc50, line 3: I get the feeling that we should not track this in version control

sdc50 (changes requested):

tethys_portal/_version.py
swainn: Removed

conda.recipe/meta.yaml
swainn: This was previously committed so we'll leave it for now.

sdc50 (approved):

PR #805, opened 2022-05-03
Remove functionReviver from PlotView Gizmo

### LINKED ISSUE #719: CSP requires unsafe-eval for plot_view
### Issue:
The Content-Security-Policy is required to include 'unsafe-eval' in the script-src section for plot_view to work. This results in a security vulnerability found during penetration testing with a CVSS of 1.4 (low severity).

### Source:
`tethys_gizmos/static/tethys_gizmos/js/plot_view.js`

```javascript
functionReviver = function(k, v) {
	if (typeof v === 'string' && v.indexOf('function') !== -1) {
		var fn;
		// Pull out the 'function()' portion of the string
		v = v.replace('function ()', '');
		v = v.replace('function()', '');

		// Create a function from the string passed in
		fn = Function(v);

		// Return the handle to the function that was created
		return fn;
	} else {
		return v;
	}
};
```

The use of `Function` is considered using eval(). We should consider refactoring this so that eval() or Function() is not required.

swainn:
Fixes #719

### DIFF
--- a/tethys_gizmos/gizmo_options/plot_view.py
+++ b/tethys_gizmos/gizmo_options/plot_view.py
@@ -198,11 +198,11 @@ class LinePlot(PlotViewBase):
                     'enabled': True,
                     'text': '{0} ({1})'.format(x_axis_title, x_axis_units)
                 },
-                'labels': {'formatter': 'function () { return this.value + " %s"; }' % x_axis_units}
+                'labels': {'units': x_axis_units}
             }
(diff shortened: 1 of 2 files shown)


msouff (approved):
This looks good to me. It turned out a lot simpler than I thought. As I mentioned, I thought there was the potential for more complex functions being past in this way, but if it's only these formatter functions then this works.

PR #841, opened 2022-07-20
Enable Tethys to be released as a PyPI package

sdc50:


sdc50:
It would be nice to have a function to create the `requirements.txt` from a conda environment. Currently it's requires some trial and error. Here is the process I just went through to creating it:

```bash
pip list --format=freeze
```

1. remove `conda`
2. remove `conda-package-handling`
3. remove `ruamel-yaml-conda`
4. change `gsconfig` to `gsconfig-py3`
5. change `psycopg2` to `psycopg2-binary`
6. change `pyproj==3.2.1` to `pyproj==3.3.1`
7. change `OWSLib==0.26.0` to `OWSLib==0.25.0`
8. comment out `tethys-dataset-services` (for now)

sdc50:
Install instructions:

1. Setup a Python 3.10 venv:
    ```bash
    conda create -n py3.10 python=3.10
    conda activate py3.10
    ```
    ```bash
    python -m venv ./venv
    ```
    ```bash
    . ./venv/bin/activate
    ```

2. From the Tethys source code install pip requirements and tethys:
    ```bash
    pip install -r requirements.txt
    pip install -e .
    ```

3. Clone tethys_dataset_services and install:
    ```bash
    git clone git@github.com:tethysplatform/tethys_dataset_services.git
    cd tethys_dataset_services
    pip install -e .
    ```

4. Start database server (use docker or another conda environment).

5. Run tethys server:
    ```bash
    tethys manage start
    ```

requirements.txt
```diff
@@ -0,0 +1,154 @@
+appdirs==1.4.4
```
swainn, line 1: I'm assuming this is just an export of the conda environment? Have you given any thought on how to maintain these dependencies?

I'd suggest that we follow a strategy like we do with conda and have the requirements.txt that is committed be unpinned as much as possible and then automatically generate the requirements.txt that is used for building the pypi package.

tethys_cli/install_commands.py
```diff
@@ -462,8 +468,14 @@ def install_command(args):
                 write_warning("Skipping package installation.")
             else:
                 if validate_schema('conda', requirements_config):  # noqa: E501
-                    conda_config = requirements_config['conda']
-                    install_packages(conda_config, update_installed=args.update_installed)
+                    if has_conda:
+                        conda_config = requirements_config['conda']
+                        install_packages(conda_config, update_installed=args.update_installed)
```
swainn, line 473: Have you looked at what it would take to refactor `install_packages()` to fallback to installing the conda packages with pip?

swainn (commented):
It would be good to update the CI with a step that builds the pypi package and uploads it automatically on release. Let's discuss more in scrum.

requirements.txt
sdc50: I had very little success in using an unpinned version of this file. Pip ends up trying to install many versions of many packages trying to figure out what will work. I described the process I went through to create it from the conda environment above. In any case we'll need to wait until `tethys_dataset_services` is fully released on PyPI, to have a complete `requirements.txt`

tethys_cli/install_commands.py
sdc50: I didn't look into that, but it's a good suggestion. I know there are many cases were packages don't have the same name on PyPi and Conda. And of course there are some packages on Conda that wouldn't be on PyPI (e.g. postgresql). But it probably would work in many cases, so it's worth trying.

requirements.txt
swainn: Follow up on this in the future. See Issue #846

swainn (approved):

sdc50:
@swainn I added a `gen` command and removed the `requirements.txt` from the repo. I propose we add this as experimental.

tethys_cli/gen_commands.py
```diff
@@ -282,6 +292,19 @@ def gen_install(args):
...
 
+def gen_requirements_txt(args):
+    write_warning('WARNING: The requirements.txt is currently only experimental.')
+    # pip list --format=freeze | sed '/conda/d'
+    output = run(['pip', 'list', '--format=freeze'], capture_output=True)
+    packages = output.stdout.decode().splitlines()
+    packages = [p for p in packages if 'conda' not in p and 'tethys-platform' not in p and 'psycopg2=' not in p]
+    context = {
+        'packages': packages,
+        'date': datetime.now().strftime('%Y-%m-%d')
```
swainn, line 303: Is there a template that goes with this? I don't see it committed.

swainn (changes requested):

swainn (approved):

PR #858, opened 2022-08-02
Update Site Settings

### LINKED ISSUE #853: Fix CSS when overriding app navigation
When the app navigation is overridden then the app icon has no left margin making it align with the left edge of the window.

### LINKED ISSUE #854: Custom styles settings are not recognized in development
When using a path for a custom style setting (e.g. `App Library CSS`), it is not recognized in a development setting. This is because it looks for the path in `settings.STATIC_ROOT`, but the `collectstatic` command has not been executed in development.

This can be fixed by using the `TethysStaticFinder` to look for the path.

### LINKED ISSUE #855: Add additional custom styles and custom template settings
For completeness I think we should add additional custom styles and custom templates settings:

image

image

- login page
- register page
- user profile
- error pages?

sdc50:
fixes icon styles when app nav is overridden - fixes #853
allow custom style settings to be found in development - fixes #854
adds additional custom style and custom template settings - fixes #855
modifies the `tethys site` command and the `portal_config.yml` schema


swainn:
Sorry @sdc50 I didn't see that this was still WIP. I updated it with master...

sdc50:
> Sorry @sdc50 I didn't see that this was still WIP. I updated it with master...

@swainn I thought that might have happened, but not a big deal. I think it's ready for review now.

tethys_apps/base/app_base.py
```diff
@@ -238,22 +238,22 @@ def register_url_maps(self):
...
+                f'Deprecation Warning: The "{self.name}" app has the `url_maps` method defined. '
+                'This method is now deprecated. '
+                'Please use the new "controller", "consumer", and "handler" decorators to register URL maps '
+                '(see docs: http://docs.tethysplatform.org/en/stable/tethys_sdk/routing.html).'
+            )
+        except AttributeError:
+            pass
+
         if self._registered_url_maps is None:
             self._registered_url_maps = self.register_url_maps()
```
swainn, line 255: Can we add an optional argument to the ``regiester_url_maps`` method that will allow passing a list of UrlMaps to initialize it with when overriding. Example usage:

```python
def register_url_maps(self):

    url_maps = [...]

    url_maps = super().register_url_maps(url_maps)

    url_maps.append(UrlMap(...))
    return url_maps
```

docs/installation/production/manual/configuration/advanced/customize.rst
```diff
@@ -8,86 +8,91 @@ Customize Portal Theme (Recommended)
...
-Home Page Template     Path to alternate Home page template (will replace Home page template entirely). The template must be located within a valid templates directory, such as in a Tethys app, Tethys extension, or Django app.
-Apps Library Template  Path to alternate Apps Library page template (will replace Apps Library page template entirely). The template must be located within a valid templates directory, such as in a Tethys app, Tethys extension, or Django app.
-====================== =================================================================================
+GENERAL_SETTINGS
+++++++++++++++++
+
+============================== ================================================================================
+Setting                        Description
+============================== ================================================================================
+SITE_TITLE                     Title of the web page that appears in browser tabs and bookmarks of the site. Default is "Tethys Portal".
```
swainn, line 17: In the Admin pages, these settings do not appear as shown in this table (upper snake case), making this a little confusing how it is written. Two suggestions:

1. Add another column to the table so it is something like this:

```rst
============================== ============================== ================================================================================
Admin Setting                  Site Setting Command           Description
============================== ============================== ================================================================================
Site Title                     SITE_TITLE                     Title of the web page that appears in browser tabs and bookmarks of the site. Default is "Tethys Portal".

...
```

2. Migrate the names of the settings in the database so that we only use the upper snake case everywhere, including in the Admin pages.

tethys_cli/gen_templates/portal_config
```diff
@@ -1,9 +1,37 @@
...
+
+############################################################################
+#
+# APPS
+#
+# This section is applied when running the "tethys install" command. Tethys
+# will first look for app settings in this file. If none are found it will
+# then look for a "services.yml" file in the install location.
+#
+#############################################################################
```
swainn, line 15: I'm not a huge fan of these bulky comment blocks, though I think the comments could be helpful. What do you think about making them a little less bulky:

```
############################################################################
# APPS
# The "tethys install" command checks this section for app settings. 
############################################################################
{{ apps }}
```

tethys_cli/site_commands.py
```diff
@@ -5,213 +5,166 @@
...
+                                  'directory, such as in a Tethys app, Tethys extension, or Django app.')
+    site_parser.add_argument('--user-settings-page-template', dest='user_settings_page_template',
+                             help='Path to alternate user settings (i.e. edit) page template (will replace settings '
+                                  'page template entirely). The template must be located within a valid templates '
+                                  'directory, such as in a Tethys app, Tethys extension, or Django app.')
+    site_parser.add_argument('-d', '--restore-defaults', dest='restore_defaults', action='store_true',
                              help='Restores the sites default values.')
-    site_parser.add_argument('-f', '--from-file', type=str, help='Load site content from portal_config.yml file.')
+    site_parser.add_argument('-f', '--from-file', action='store_true', default=False,
+                             help='Load site content from portal_config.yml file.')
```
swainn, line 144: Our Docker build will need to be updated in at least two places to account for these changes:

1. The `tethyscore.sls` script will need to be updated (overhauled) accordingly. See: https://github.com/tethysplatform/tethys/blob/master/docker/salt/tethyscore.sls#L157
2. The `Dockerfile` environment variables. See: https://github.com/tethysplatform/tethys/blob/master/Dockerfile#L57

swainn (changes requested):
@sdc50 this is a lot of work! All for the better of Tethys. I have a few suggestions. See comments.

tethys_cli/gen_commands.py
```diff
@@ -186,11 +187,27 @@ def gen_nginx_service(args):
...
+    tethys_portal_settings.setdefault('settings', {'SECRET_KEY': generate_secret_key()})
+    tethys_portal_settings.setdefault('site_settings', {category: {} for category in SITE_SETTING_CATEGORIES})
+
+    try:
+        tethys_portal_settings.update(args.tethys_portal_settings)
+    except AttributeError:
 
-    context = {'SECRET_KEY': generate_secret_key()}
+        write_info('A Tethys Portal configuration file is being generated. '
+                   'Please review the file and fill in the appropriate settings.')
```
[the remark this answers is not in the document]
sdc50: Because the path gets output elsewhere. Having the path here was just a duplicate.

tethys_cli/gen_templates/portal_config
sdc50: I like having them be less bulky. My main concern is that it was confusing that each section of the `portal_config.yml` was applied at different times and in different ways. I was hoping having the comments in the file itself would help clarify how each section is used. I'll adjust them to be more concise.

tethys_cli/gen_commands.py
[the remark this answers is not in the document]
sdc50: Previous Message:
image

New Message:
image

tethys_apps/base/app_base.py
sdc50: @swainn What are you expecting the `register_url_maps` to do with the list of `UrlMap`s that you pass to it? In you're example you pass some in to the `super` call, and then also append some afterward. So I'm assuming there is something different about the once you pass vs the ones you append after.

docs/installation/production/manual/configuration/advanced/customize.rst
sdc50: I don't think we should migrate the names of the settings in the DB. I like the more readable names as they are. 

I think the setting names in the `portal_config.yml` file were made all caps so they would follow the convention of the Django settings. Even if we changed them to to follow the title case of the DB settings, we would still need to have underscores. I think just having a table that shows the correlation would be good, but we may need to have different tables in different places.

For example in the Production Installation guide these settings are only discussed in terms of the Admin Pages, so we probably don't need to include the `portal_config` key. However, on the customization page it would be good be more complete.

Also, in the suggested table you have a column called `Site Setting Command`. That column would more accurately be `Portal Config Yaml Key`. The site setting command for that setting would actually be called like this:

```bash
tethys site --site-title "Title of Site"
```

We could include a `Site Setting Command` column in the table for completeness:

```
============================== ============================== ============================== ================================================================================
Admin Setting                  Portal Config Yaml Key         Site Setting Command           Description
============================== ============================== ============================== ================================================================================
Site Title                     SITE_TITLE                     --site-title                   Title of the web page that appears in browser tabs and bookmarks of the site. Default is "Tethys Portal".

...
```

@swainn what are your thoughts?

tethys_cli/gen_commands.py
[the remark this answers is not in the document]
swainn: Thank you for the clarification.

tethys_apps/base/app_base.py
swainn: In AGWA we ran into an edge case. The app defines no URL endpoints itself with the controller decorator but instead uses ATCore to generate its UrlMaps. The endpoint used as the home view is included in these generated URLs. The problem is that the `register_url_maps` method looks for a controller matching the `index` property and fails if it is not found.

We were able to work around this issue by not calling the super method in our override. However, if AGWA does end up implementing its own controllers in the future with the controller decorator, we'd have no way to load them, since if we call the super method, it errors out.

My suggestion is to allow seeding the `register_url_maps` method with manually created UrlMaps. Another alternative I just thought of is that we have an optional parameter for specifying only the home UrlMap for cases when it is manually defined and not decorated.

docs/installation/production/manual/configuration/advanced/customize.rst
swainn: I like your table showing the relationship between the admin settings, portal yaml key, and site setting option. What if instead of having the table in multiple places, we put it in one place and then place `ref` links to it in the other places it is needed?

tethys_apps/base/app_base.py
sdc50: I think that you mentioned that your home controller is actually provided by an extension, so I'm guessing that's the reason for wanting to pass a list of maps? And adding extra controller modules via the app class `controller_modules` doesn't work?

Here is the actual implementation of the `register_url_maps`:

```python
        from .controller import register_controllers

        controller_modules = [f'{self.package_namespace}.{self.package}.{module_name}' for module_name in
                              set(DEFAULT_CONTROLLER_MODULES + self.controller_modules)]
        return register_controllers(
            root_url=self.root_url,
            modules=controller_modules,
            index=self.index,
            catch_all=self.catch_all,
        )
``` 

The only way for there to be any value in passing anything to this method would be if we also pass it through to the `register_controllers` method. The `register_cpontrollers` method actually creates 'UrlMaps` so it may be best to just pass in the `UrlMap` kwargs.

Implementation would looks something like this:


App Class Override:

```python
def register_url_maps(self):

    controller_kwargs = [
        dict(
            name='url_name',
            url='final_url',
            controller='controller',
            protocol='protocol',
            regex='regex',
            handler='handler',
            handler_type='handler_type'
        ),
    ]


    url_maps = super().register_url_maps(controller_kwargs)

    url_maps.append(UrlMap(...))
    return url_maps
```

app_base .py

```python
        from .controller import register_controllers

        controller_modules = [f'{self.package_namespace}.{self.package}.{module_name}' for module_name in
                              set(DEFAULT_CONTROLLER_MODULES + self.controller_modules)]
        return register_controllers(
            root_url=self.root_url,
            modules=controller_modules,
            index=self.index,
            catch_all=self.catch_all,
            extra_controller_kwargs=controller_kwargs,
        )
``` 

controllers.py

```python
def register_controllers(
        root_url: str,
        modules: Union[str, list, tuple],
        index: str = None,
        catch_all: str = '',
        extra_controller_kwargs: list = None,
) -> list:
    global app_controllers_list
    ...
    app_controllers_list = list()
    for module in all_modules:
        importlib.reload(module)  # load again to register controllers

    app_controllers_list.extend(extra_controller_kwargs)

    names = dict()
    for kwargs in app_controllers_list:
        name = kwargs['name']

    ...
```

sdc50: @swainn Sorry, I commented before refreshing and seeing your response. It seems like the easiest thing to do in your example would just be to have an option to disable the auto-register from the home controller.  The `register_controllers` already accepts an `index` argument. If it's `None` then it will skip over any of the home controller logic, so you could just do this:

```python
def register_url_maps(self):

    url_maps = super().register_url_maps(index=None)

    url_maps.extend(url_maps_from_extension)
    return url_maps
```

Would that work?

docs/installation/production/manual/configuration/advanced/customize.rst
sdc50: These tables would go in two places:

http://docs.tethysplatform.org/en/stable/installation/production/manual/configuration/advanced/customize.html

and 

http://docs.tethysplatform.org/en/stable/tethys_portal/configuration.html#site-settings

If we only had it in one place I would say it should go in the second place. Is that okay:

image

swainn: Yes.

sdc50: @swainn It looks like a similar table is also in the docker reference: http://docs.tethysplatform.org/en/stable/installation/production/docker/tethys_docker_reference.html#tethys-portal-site-settings

Do you think I should I just reference the table here too, and just explain that the docker environment variable will be the same as the portal config key?

swainn: Sounds good to me.

tethys_apps/base/app_base.py
swainn: I think so... I'm trying to remember how Tethys knows which URL to use as the home URL and what role the register_url_maps function has in that.

tethys_apps/base/app_base.py
```diff
@@ -178,10 +178,13 @@ def urlpattern(suffix=""):
...
+    def register_url_maps(self, set_index=True):
         """
         Only override this method to manually define or extend the URL Maps for your app. Your ``UrlMap`` objects must be created from a ``UrlMap`` class that is bound to the ``root_url`` of your app. Use the ``url_map_maker()`` function to create the bound ``UrlMap`` class. Starting in Tethys 3.0, the ``WebSocket`` protocol is supported along with the ``HTTP`` protocol. To create a ``WebSocket UrlMap``, follow the same pattern used for the ``HTTP`` protocol. In addition, provide a ``Consumer`` path in the controllers parameter as well as a ``WebSocket`` string value for the new protocol parameter for the ``WebSocket UrlMap``. Alternatively, Bokeh Server can also be integrated into Tethys using ``Django Channels`` and ``Websockets``. Tethys will automatically set these up for you if a ``handler`` and ``handler_type`` parameters are provided as part of the ``UrlMap``.
 
+        Args:
+            set_index: Whether to pass the ``index`` argument to ``register_controllers``. If ``True`` (default) then ``self.index` will be passed. Otherwise ``None`` will be passed as the ``index`` argument.
```
swainn, line 186: @sdc50 Not being familiar with the implementation of register_url_maps, this explanation doesn't help much. What is the result of index being passed as `None`?

swainn: In other words, why would I set `set_index` to `False`?

tethys_apps/base/app_base.py
```diff
@@ -178,10 +178,13 @@ def urlpattern(suffix=""):
...
+    def register_url_maps(self, set_index=True):
         """
         Only override this method to manually define or extend the URL Maps for your app. Your ``UrlMap`` objects must be created from a ``UrlMap`` class that is bound to the ``root_url`` of your app. Use the ``url_map_maker()`` function to create the bound ``UrlMap`` class. Starting in Tethys 3.0, the ``WebSocket`` protocol is supported along with the ``HTTP`` protocol. To create a ``WebSocket UrlMap``, follow the same pattern used for the ``HTTP`` protocol. In addition, provide a ``Consumer`` path in the controllers parameter as well as a ``WebSocket`` string value for the new protocol parameter for the ``WebSocket UrlMap``. Alternatively, Bokeh Server can also be integrated into Tethys using ``Django Channels`` and ``Websockets``. Tethys will automatically set these up for you if a ``handler`` and ``handler_type`` parameters are provided as part of the ``UrlMap``.
 
+        Args:
+            set_index: If `False` then the index controller will not be configured/validated automatically, and it is left to the user to ensure that a controller name that matches `self.index` is configured.
```
sdc50, line 186: @swainn is this a better explanation of what `set_index` does?

tethys_apps/base/app_base.py
sdc50: see comment below

swainn (approved):

PR #861, opened 2022-08-09
Fix condor logs

### LINKED ISSUE #617: Job logs work while job is running but not after job is done [bug]
Once the job is completed, an exception is thrown when trying to access the logs from the jobs table.

@nathan was able to reproduce it. @sdc50 have you seen this?

sdc50:
fix condor logs fixes #617

### DIFF
--- a/tests/unit_tests/test_tethys_compute/test_models/test_CondorBase.py
+++ b/tests/unit_tests/test_tethys_compute/test_models/test_CondorBase.py
@@ -189,97 +189,92 @@ class CondorBaseTest(TethysTestCase):
         self.assertEqual(expected_result, log_content)
         mock_log.exception.assert_called_once()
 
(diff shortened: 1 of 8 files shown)


sdc50:
@msouff do you have the ability to test this out easily? From what I understand of the issue this should fix things, but I don't have a condor environment set up to easily test it out.

swainn:
@msouff you could check out this PR branch on the AGWA machine that you are working on for the Tethys 4 upgrade.

msouff:
> @msouff you could check out this PR branch on the AGWA machine that you are working on for the Tethys 4 upgrade.

Yes, I can try that. I don't think we ever remove local files from workspaces unless the job is deleted but I wasn't confident enough to answer. I'll it test tomorrow and report back.

swainn (approved):

PR #866, opened 2022-08-15
Install yml tethys version

msouff:
added tethys_version to install.yml template

### DIFF
--- a/docs/installation/application.rst
+++ b/docs/installation/application.rst
@@ -67,7 +67,7 @@ The install command uses three configuration files:
 
 .. _tethys_install_yml:
 
-install.yml 
+install.yml
 -----------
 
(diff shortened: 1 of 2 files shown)


swainn:
@sdc50 , we want to add a `tethys_version` to the `install.yml` as metadata for the App Store to use to know what versions of Tethys an app can be installed on. We considered listing tethys-platform as a dependency in the conda deps, but were concerned it may result in unintended upgrades of portals. So our current approach is to list it as a separate parameter. Please review and provide your feedback.

sdc50 (commented):
Is this just intended to be metadata for the user's information? It looks like there is not any code that actually reads the tethys_version value.

swainn:
> Is this just intended to be metadata for the user's information? It looks like there is not any code that actually reads the tethys_version value.

Yes. For now.

PR #867, opened 2022-08-24
Add get_workspace utility functions that enforce quotas

swainn:
Adds `get_app_workspace` and `get_user_workspace` functions that can be used when the decorator cannot be used. 

- These functions will enforce Workspace Quotas when enabled.
- They raise an Assertion Error when the Quota is not met.

### DIFF
--- a/tethys_apps/base/workspace.py
+++ b/tethys_apps/base/workspace.py
@@ -211,29 +211,6 @@ def _get_user_workspace(app_class, user_or_request):
 
     Returns:
       tethys_apps.base.TethysWorkspace: An object representing the workspace.
-
-    **Example:**
-
-    ::
-
-        import os
-        from my_first_app.app import MyFirstApp as app
-
-        def a_controller(request):
-            \"""
-            Example controller that uses get_user_workspace() method.
-            \"""
-            # Retrieve the workspace
-            user_workspace = app.get_user_workspace(request.user)
-            new_file_path = os.path.join(user_workspace.path, 'new_file.txt')
-
-            with open(new_file_path, 'w') as a_file:
-                a_file.write('...')
-
-            context = {}
-
-            return render(request, 'my_first_app/template.html', context)
-
     """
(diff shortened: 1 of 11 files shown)


tethys_apps/base/workspace.py
```diff
@@ -254,6 +256,42 @@ def a_controller(request):
...
+    elif isinstance(request_or_user, User):
+        user = request_or_user
+    else:
+        raise ValueError(f'Argument "request_or_user" must be of type HttpRequest or User: '
+                         f'"{type(request_or_user)}" given.')
+
+    assert passes_quota(user, 'user_workspace_quota')
+
+    # Get the active app
+    app = get_active_app(request_or_user, get_class=True)
```
msouff, line 288: Have you tested the behavior if a User is passed? I'm not sure get_active_app accept anything other than a request since it needs the url to figure out what the active app is.

sdc50 (commented):
What if we just add these methods back to the app class? 

```python
from .app import MyApp as app

app.get_app_workspace()
app.get_user_workspace(user_or_request)
```

tethys_apps/base/workspace.py
swainn: Good catch. I've refactored this and tested it.

swainn:
> What if we just add these methods back to the app class?
> 
> ```python
> from .app import MyApp as app
> 
> app.get_app_workspace()
> app.get_user_workspace(user_or_request)
> ```

Done.

swainn:
@sdc50 this is ready for review assuming tests pass.

swainn:
Tests are failing for an unrelated reason. The tests that fail are running into an infinite recursion issue with mock. They pass in my environment which has python 3.10.4. The CI environment is resolving to 3.10.6.

swainn:
#870 fixes the failing tests in this PR

sdc50 (approved):

PR #876, opened 2022-10-01
Shall we Black our Codebase?

swainn:
This PR includes the following:

- Run [Black Python code formatter](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) on Tethys codebase
- Update Flake8 configuration per [Black recommendations](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length)
- Enable the Black GitHub action in our CI (see below)

Note: as a side-effect of this change, the max line-length was changed to a soft 88 characters.

It's probably not prudent to look at every file in this PR (since it is pretty much every file in the repo), but it is useful to look at code you are familiar with to see how the formatting was changed.


sdc50:
@swainn I agree with most of the changes that Black makes. I think it sometimes makes things less readable (i.e. removing blank lines in the `controller` decorator signature, but I'm willing to live with a few disagreements.

sdc50 (approved):

PR #884, opened 2022-11-18
Expand apache/nginx options

sdc50:


Dockerfile
```diff
@@ -175,7 +175,7 @@ RUN apt-get -y remove gcc \
 # CONFIGURE  ENVIRONMENT#
 #########################
 ENV PATH ${CONDA_HOME}/miniconda/envs/tethys/bin:$PATH 
-VOLUME ["${TETHYS_HOME}/workspaces", "${TETHYS_HOME}/keys"]
+VOLUME ["${WORKSPACE_ROOT}", "${TETHYS_HOME}/keys"]
```
sdc50, line 178: @swainn I was going to ask you why we have this specific directories mapped as volumes. It seems like we should map the whole `TETHYS_PERSIST` dir.

sdc50:
@swainn I also wanted to discuss having a single `tethys-docker` repository that has build args to configure the build to use either the `micromamba` base image or the `redhat/ubi` base image. Additionally, it could be configured to use either `Nginx` or `Apache`. That way we could just maintain one `Dockerfile` but be able to build several tethys images from it.

Dockerfile
swainn: Remind me are these directories in the `TETHYS_PERSIST` directory? If so, then yes. We should do that. I think that's how we configure our k8s deployments anyway.

docker/salt/tethyscore.sls
```diff
@@ -27,6 +27,7 @@
 {% set TETHYS_PERSIST = salt['environ.get']('TETHYS_PERSIST') %}
 {% set STATIC_ROOT = salt['environ.get']('STATIC_ROOT') %}
 {% set WORKSPACE_ROOT = salt['environ.get']('WORKSPACE_ROOT') %}
+{% set BYPASS_TETHYS_HOME_PAGE = salt['environ.get']('BYPASS_TETHYS_HOME_PAGE') %}
```
swainn, line 30: Is this used anywhere in this file?

swainn (commented):
See comments.

swainn (approved):

Dockerfile
sdc50: The default `WORKSPCE_ROOT` is in `TETHYS_PERSISIT`.  The `${TETHYS_HOME}/keys` is obviously in `TETHYS_HOME`, but it's not used anywhere.

docker/salt/tethyscore.sls
sdc50: Oops! It looks like I had already added it in my last PR.

swainn (approved):

PR #890, opened 2022-12-06
Remove array fields

sdc50:
Replaces ArrayFields with JSONFields. 
Update documentation on app testing.
Fix annoyances with app settings:

- Ensure that new settings are synced when syncing an app with the database
- Ensure the obsolete settings are removed when syncing an app
- Prompt user if app should be removed from database when there is an error when loading in `DEBUG` mode.

PR #898, opened 2022-12-22
Fixed dynamic settings sync bug

swainn:
Fixes a bug with app setting syncing introduced with #890.

- Persistent store database settings that were created dynamically using the `app.create_persistent_store()` method were removed every time Tethys reloaded.
- This is because dynamically created settings are not listed in the app.py and were thus being treated as obsolete settings.
- This PR makes it so that settings with a `dynamic` attribute that is `True` will be ignored when removing obsolete settings.


shawncrawley (approved):
Looks good to me. Approved.

sdc50 (approved):

PR #912, opened 2023-01-26
fix ensure_oauth logic

sdc50:
The `get_access_token` method has the same logic in it already that I had implemented in the `ensure_oauth2` decorator. In addition to being redundant it was also not handling an error if the token is not valid.


swainn (approved):
Approved after tests and other checks pass.

swainn (approved):

PR #914, opened 2023-01-30
Fix for workflows when you add a point and go back

douggallup:
Fix the javascript so that on a workflow that adds points, if you go to the next page, but then go back to the previous page (where the point was added), you get the points displayed again.


swainn:
More generally, this fixes an issue when the initial drawing features is given as JSON objects rather than an JSON string like it expects. The MapView gizmo will now handle both cases.

swainn (approved):

PR #927, opened 2023-03-01
Enable other Databases

sdc50:
Enable other databases (e.g. sqlite) to be used as the primary Tethys database

Note that this PR does not change the default database configuration. It just enables other databases to be configured, including SQLite. In the case of SQLite it provides some support comparable to Postgres with the `tethys db` commands (e.g. `tethys db configure` and `tethys db purge`).


environment.yml
```diff
@@ -82,7 +81,7 @@ dependencies:
   - flake8-bugbear
   - git
   - setuptools_scm
-  - openssl<3.0.0 # Pin until openssl issue is resolved by conda
+  - openssl
```
swainn, line 84: I assume this means they resolved the issue?

tethys_cli/db_commands.py
```diff
@@ -219,8 +220,18 @@ def purge_db_server(**kwargs):
         **kwargs: processed key word arguments from commandline
     """
     kwargs["exit_on_error"] = False
-    stop_db_server(**kwargs)
+    delete_function = None
     if kwargs["db_dir"]:
+        stop_db_server(**kwargs)
+        delete_function = partial(shutil.rmtree, kwargs["db_dir"])
+    elif kwargs["db_engine"].endswith("sqlite3"):
```
swainn, line 227: I suggest using `"sqlite" in kwargs["db_engine"]` to future-proof ourselves.

tethys_cli/db_commands.py
```diff
@@ -470,15 +482,22 @@ def process_args(args):
...
-        if not Path(db_dir).is_absolute():
-            db_dir = Path(get_tethys_home_dir()) / db_dir
+        db_dir = relative_to_tethys_home(db_dir)
+
+    db_engine = db_settings.get("ENGINE")
+    if not db_engine.endswith("postgresql"):
+        if args.command == "create":
+            raise RuntimeError(
+                f"The tethys db create command can only be used with PostgreSQL databases."
+            )
```
swainn, line 500: I'm assuming this is because the method for creating the database is postgresql specific at this moment? In the future it would be good to support creating the database in other non-sqlite dbs.

tethys_portal/settings.py
```diff
@@ -20,28 +20,28 @@
...
 from tethys_cli.gen_commands import generate_secret_key
 
 from bokeh.settings import settings as bokeh_settings, bokehjsdir
 
 log = logging.getLogger(__name__)
 this_module = sys.modules[__name__]
 
-BASE_DIR = os.path.dirname(__file__)
-TETHYS_HOME = get_tethys_home_dir()
+BASE_DIR = Path(__file__).parent
```
swainn, line 40: I'm a big fan of this change. Now could you do it for the rest of the repo? ;)

tests/unit_tests/test_tethys_apps/test_models/test_PersistentStoreDatabaseSetting.py
```diff
@@ -20,16 +20,12 @@ class PersistentStoreDatabaseSettingTests(TethysTestCase):
...
-                "PASSWORD": "pass",
-                "HOST": "localhost",
-                "PORT": "5435",
-            }
+        self.conn = {
+            "USER": "tethys_super",
+            "PASSWORD": "pass",
+            "HOST": "localhost",
+            "PORT": "5435",
+        }
```
swainn, line 28: Why this change?

swainn (commented):
See comments

environment.yml
sdc50: Yes, I tested it unpinned, and it worked in both a Python 3.11 and 3.10 environments.

tethys_cli/db_commands.py
sdc50: That's right. The code that we have to create and setup the database is specific to Postgres.

tests/unit_tests/test_tethys_apps/test_models/test_PersistentStoreDatabaseSetting.py
sdc50: I'm not sure why it was using the actual database configuration for these tests in the first place. It seems like for the testing environment we want to control the inputs. 

In any case, when using sqlite then the database configuration won't have any this information, so getting the actual database config will cause the tests to fail.

tethys_portal/settings.py
sdc50: Looks like over 100 changes. I'm happy to do it, but do you think I should do it in a separate PR?

swainn: I was sort of joking, but hey if you want to do it that would be awesome. Separate PR for sure.

tethys_cli/gen_commands.py
```diff
@@ -465,7 +465,7 @@ def gen_requirements_txt(args):
     packages = [
         p
         for p in packages
-        if "conda" not in p and "tethys-platform" not in p and "psycopg2=" not in p
+        if all([s not in p for s in ["conda", "tethys-platform", "psycopg2=", "tethysapp."]])
```
swainn, line 468: Might as well add "tethysext." as well.

swainn, line 468:
```suggestion
        if all([s not in p for s in ["conda", "tethys-platform", "psycopg2=", "tethysapp.", "tethysext."]])
```

swainn (approved):
Conditionally approved - get checks passing before merging.

Issue #925, opened 2023-02-28
Make Sqlite the default database for development


sdc50:
I propose that we make sqlite3 the default database engine for Tethys. This will make a simpler development environment and eliminate any confusion with whether or not  the database server is running, or if there multiple databases that use the same port.

I think this will also give Tethys Portal managers more flexibility in deciding what database to set up in production.

To prevent a breaking change I would add code that detects if a postgres database is configured and preserve the postgres `ENGINE` configuration with a warning that it will change in the future. 

One concern is that there could be possible differences with how SQLAlchemy interacts with a sqlite database vs PostgreSQL or other databases. I'll need to research this to ensure that there will not be potential pitfalls for app developers.

sdc50:
@swainn Here is the message that would be displayed if a postgres database configuration is detected:

![image](https://user-images.githubusercontent.com/6454472/221987483-fde3bb3f-35a4-4421-8fe2-19456029c257.png)

sdc50:
If no database configuration exists, but the default `psql` database dir exists in `TETHYS_HOME` then the message would be slightly different:

![image](https://user-images.githubusercontent.com/6454472/221987802-916702ea-76ec-4c05-9837-c35d2fcc7962.png)

swainn:
I have a few additional suggestions:
1. Provide various data migration command for `tethys db` that would allow users to migrate data from one type of DB to another. Potential use cases include:
    -   Someone wanting to move their development environment from an existing PostgreSQL database to the new default SQLite database.
    -   Someone who has outgrown the SQLite database for development to move to a new engine.
2. Update the documentation for database settings:
    -  Tethys Portal Configuration documentation to include the `DATABASES.default.ENGINE` setting.
    -  Manual Production Deployment documentation to include warnings about using SQLite in production.
    -  Update the Manual Production Deployment documentation to include instructions for configuring various DB backends in production (PostgreSQL, MySQL, Microsoft SQL, etc.).

swainn:
What do you think about changing the CI to run the tests multiple times with different db backends to help us get some measure of assurance that it works for each of them? I'm thinking sqlite, postgresql, and mysql at least. If it's possible to get Microsoft SQL instances in GitHub actions without needing to license it, then we could do that one too.

sdc50:
[...]

PR #928, opened 2023-03-01
Make sqlite default db

sdc50:
Changes the default database configuration in settings to use SQLite instead of Postgres. 

Provides backwards compatibility code to detect if a default Postgres database was already configured to preserves the `postgresql` engine in that case.


sdc50:
Remaining work for this PR:

- [ ] Add data migration command/instructions to switch from one type of DB to another
- [ ] Update the documentation for database settings
- [ ] Update deployment documentation

see: https://github.com/tethysplatform/tethys/issues/925#issuecomment-1450384795

sdc50:
@swainn  I think this is ready to go. As for the production documentation I think that it is out of scope to write a detailed guide for setting up other types of databases, and if we ever did I would want to throughly test them in a production environment first. What I did include is an explanation that other database backends are possible and referred to the Django Database docs for more details.

sdc50:
Also, it looks like the most recent version of `python-social-auth` was breaking our tests. As far as I can tell it just dropped support for a very outdated way of using middleware, but our tests were still calling it with the old style (not passing in a `get_request` argument).

swainn:
> Also, it looks like the most recent version of `python-social-auth` was breaking our tests. As far as I can tell it just dropped support for a very outdated way of using middleware, but our tests were still calling it with the old style (not passing in a `get_request` argument).

Is it only our tests that are using the old way? Is the code up-to-date?

tests/unit_tests/test_tethys_portal/test_middleware.py
```diff
@@ -22,13 +22,14 @@ def tearDown(self):
...
     ):
+        mock_get_response = mock.MagicMock()
         mock_request = mock.MagicMock()
         mock_exception = mock.MagicMock()
         mock_hasattr.return_value = True
         mock_isinstance.return_value = True
         mock_request.user.is_anonymous = True
 
-        obj = TethysSocialAuthExceptionMiddleware()
+        obj = TethysSocialAuthExceptionMiddleware(mock_get_response)
```
swainn, line 32: Is there a corresponding change we need to make in the `TethysSocialAuthExceptionMiddleware`?

tethys_cli/db_commands.py
```diff
@@ -33,18 +33,22 @@ def add_db_parser(subparsers):
         "command",
         help="Performs operations on the Tethys Database.\n"
         "  * init - Creates a new, locally running PostgreSQL database server.\n"
-        "  * start - Starts the local database server.\n"
-        "  * stop - Stops the local database server.\n"
-        "  * status - Gets the current status of the local database server.\n"
-        "  * create - Creates the Tethys Portal database on the database server configured "
+        "  * start - Starts the local (PostgreSQL) database server.\n"
+        "  * stop - Stops the local (PostgreSQL) database server.\n"
+        "  * status - Gets the current status of the local (PostgreSQL) database server.\n"
```
swainn, line 38: "the local (Postgresql)" -> "a local Postgresql"

swainn (commented):
Do we need to add sqlite3 to the environment.yml? Are we planning to take out the postgresql dependencies and add instructions for installing them when you want to use postgresql?

tethys_cli/db_commands.py
swainn: *PostgreSQL

sdc50:
> Do we need to add sqlite3 to the environment.yml? Are we planning to take out the postgresql dependencies and add instructions for installing them when you want to use postgresql?

sqlite3 is native to Python so no dependencies needed. I wasn't planning to remove the PostgreSQL dependencies. I think they are still handy to have available to test production environments.

tests/unit_tests/test_tethys_portal/test_middleware.py
sdc50: No, as far as I can tell it was already compatible with the newer way of calling it (passing in the get_response). The latest change just dropped support for the old way of calling it (without the get_response arg). The only change necessary was to update our tests to use the new style.

tethys_cli/db_commands.py
```diff
@@ -33,18 +33,22 @@ def add_db_parser(subparsers):
         "command",
         help="Performs operations on the Tethys Database.\n"
         "  * init - Creates a new, locally running PostgreSQL database server.\n"
-        "  * start - Starts the local database server.\n"
-        "  * stop - Stops the local database server.\n"
-        "  * status - Gets the current status of the local database server.\n"
-        "  * create - Creates the Tethys Portal database on the database server configured "
+        "  * start - Starts the local (PostgreSQL) database server.\n"
```
sdc50, line 36:
```suggestion
        "  * start - Starts a local PostgreSQL database server.\n"
```

tethys_cli/db_commands.py
```diff
@@ -33,18 +33,22 @@ def add_db_parser(subparsers):
...
+        "  * stop - Stops the local (PostgreSQL) database server.\n"
```
sdc50, line 37:
```suggestion
        "  * stop - Stops a local PostgreSQL database server.\n"
```

tethys_cli/db_commands.py
sdc50, line 38:
```suggestion
        "  * status - Gets the current status of a local PostgreSQL database server.\n"
```

sdc50:
> Is it only our tests that are using the old way? Is the code up-to-date?

The `TethysSocialAuthExceptionMiddleware` class is never actually instantiated in our code.  It's added as a middleware, so I think it's only instantiated by Django.

swainn (approved):

PR #923, opened 2023-02-18
JSON and Secrets Custom Settings

romer8:
The old CustomSettings  is divided into three different custom settings. Simple, JSON and Secrets Custom settings.  In addition two new command lines functions are added one for generating salt strings for custom settings, and one to create a new secrets.yml file containing salt strings

PR #929, opened 2023-03-02
Adds JSON and Secret Custom Settings

swainn:
Moved @romer8 's branch into Tethys repo. See #923 for review comments.

image


swainn:
@sdc50 and @shawncrawley this is ready for review. We're waiting on conda-forge to merge conda-forge/staged-recipes#22096, so the tests are failing.

tethys_apps/migrations/0004_auto_20230223_1731.py
```diff
@@ -0,0 +1,107 @@
+# Generated by Django 3.2.16 on 2023-02-23 17:31
```
sdc50, line 1: I would recommend deleting all of these migrations, rebasing this branch, setting up a new DB, and then re-create a single migration.

sdc50 (changes requested):

tethys_apps/migrations/0004_auto_20230223_1731.py
swainn: Done

shawncrawley (approved):
This all looks good from what I could gather. I largely skimmed the test files, but went slowly and in-depth into the actual implementation. Good stuff.

PR #939, opened 2023-04-10
Update the docker command to use unidata/thredds-docker:4.6.20

swainn:
The `unidata/thredds-docker:4.6.20-SNAPSHOT` image has been removed now that 4.6.20 is out.

Fixes #929


shawncrawley (approved):
Both the most thorough and the quickest review I've done. Approved.

Issue #933, opened 2023-03-22
[FEATURE] Add conda alias for micromamba in Docker image


swainn:
**Is your feature request related to a problem? Please describe.**
When building Dockerfiles derived from the Tethys Docker image, developers run into issues running conda commands. This is because the new version of the Tethys Docker image is built with micromamba, which has some fundamental differences. For example there is no base environment and the command is `micromamba` instead of `conda`.

**Describe the solution you'd like**
The `micromamba` command has the same API as the `conda` command, so I recommend we add a `conda` alias to the `micromamba` command to help alleviate some of these issues. We could also add something like a `CONDA_EXE` environment variable.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #942, opened 2023-04-19
Various Improvements for Deployment

swainn:
This PR provides the following fixes/improvements for the Docker deployment:
* Adds the `NGINX_PORT` ENV for the Dockerfile to allow setting the port NGINX binds to
* In salt scripts, surrounds values used in commands with quotes.

Addresses the following other issues:
* Addresses a Django 4 deprecation warning to use gettext_lazy instead of ugettext_lazy.
* Fixes a bug with the check for the "migrate" arg before harvesting.
* Implements #933


swainn:
@sdc50 did you do any work with making the salt scripts more dynamic/intelligent when you were working on the redhat dockefile? i.e.: they don't run only on the first startup.

tethys_apps/apps.py
```diff
@@ -21,7 +21,7 @@ def ready(self):
         """
         Startup method for Tethys Apps django app.
         """
-        if sys.argv[1] != "migrate":
+        if len(sys.argv) > 1 and sys.argv[1] != "migrate":
```
sdc50, line 24: Did you run into a case where this was called without `sys.argv > 1`?

sdc50:
> @sdc50 did you do any work with making the salt scripts more dynamic/intelligent when you were working on the redhat dockefile? i.e.: they don't run only on the first startup.

I've made a few changes, mainly adding additional environment variables to act as flags for certain steps. I'm planning to merge the docker files together at some point.

Issue #907, opened 2023-01-05
Add port argument to the tethys schedulers create-condor command


swainn:
**Is your feature request related to a problem? Please describe.**
Cannot specify the port when creating a condor scheduler with the `tethys schedulers` command.

**Describe the solution you'd like**
Allow setting the port when creating a condor scheduler from the command line.

**Describe alternatives you've considered**
N/A

**Additional context**
See: https://github.dev/tethysplatform/tethys/blob/6f700333388ffb5e76b0225c2cd8c2eb00c97f63/tethys_cli/scheduler_commands.py#L17

PR #947, opened 2023-05-09
Add Port Option to tethys scheduler create-condor CLI command

swainn:
Implements #907

* Adds `-o/--port` option to the `tethys scaffold create-condor` command
* Includes the `port` attribute in the `tethys scaffold list -t condor` command output


shawncrawley (approved):
Good to go.

PR #949, opened 2023-05-10
Fix Site Settings

### LINKED ISSUE #930: Remove "Call to Action" properties from Home Page settings
It appears that the `Call to Action` and `Call to Action Button` elements have been removed from the home page. The properties that populated the text for those elements are still in the database and should be removed.

### LINKED ISSUE #934: Some Custom Styles Settings are not being used
I can't find anywhere where the custom styles `user_base_css` or `accounts_base_css` are being applied. It seems like these are missing from the base templates.

Also, I don't think we even have a user "base" template per se. The `profile` template extends the global base, and the other templates extend the `profile` template. So I guess the `profile` template would be considered the `user_base` template, and should be where the `user_base_css` is applied.

sdc50:
remove "Call to Action" settings
create a user base template
add custom CSS to user base and accounts base templates
restructure user templates to inherit from the new user base

fixes #930 fixes #934


swainn (approved):
@sdc50 this looks good to me though I wasn't able to pull it up and test it. Any pages you didn't test manually?

PR #953, opened 2023-05-12
Fix Packaging Options

sdc50:
modify package discovery options to find sub-packages and package data


swainn (approved):
Provided checks pass

PR #956, opened 2023-05-30
Fix for Feature Selection Hang on MapView Gizmo

swainn:
- The WMS layer feature selection feature would hang when more than 2 layers were involved in the selection. This was due to nested loops using the same looping variable (`i`). This PR fixes the issue by renaming the variable in the outer loop.
- This PR also updates deprecated `autoPan` properties of `ol.Overlays`.


shawncrawley (approved):
Good to go.

PR #958, opened 2023-05-31
Hotfix - remove args and kwargs from default Bokeh controller

sdc50:
I'm not sure why I added these, but I can't think of how they would ever be used and it turns out that it messes up the URL registration and requires that `args` and `kwargs` be passed in as URL variables.


swainn (approved):

swainn:
@sdc50 can you also cherry-pick onto branch of main and submit PR to main?

PR #962, opened 2023-06-01
Hotfix - remove only/all settings

sdc50:
fix the `sync_settings` code so that if all settings are removed the logic that deletes them from the DB will still execute.


sdc50:
replaced by #964

PR #967, opened 2023-06-16
Optional Dependencies

swainn:
This PR makes most Python dependencies optional.

TODO:

- [x] Tests and Coverage
- [x] Linting and Black
- [x] Documentation include instructions for dependency installation
- [x] Test manual production deployment
- [x] Update `tethys gen metayaml` to generate minimal and full versions of the environment
- [x] Update CI to build `micro-tethys-platform` package in addition to the "normal/full" build


swainn:
Add instructions to the documentation for installing dependencies for the following features:

- [ ]

docs/tethys_sdk/gizmos/bokeh_view.rst
```diff
@@ -2,7 +2,24 @@
...
+
+   .. code-block:: bash
+
+      # conda: conda-forge channel strongly recommended
+      conda install -c conda-forge "bokeh<3"
+
+      # pip
+      pip install "bokeh<3"
+
+   **Don't Forget**: If you end up using this gizmo in your app, add `bokeh` as a requirement to your `install.yml`.
```
swainn, line 19: @sdc50 I took a first pass at a dependency install tip box in docs for optional dependencies. What do you think of this?

swainn: Looks like this:

image

sdc50: This looks great! 

As a side note I think we can take off the pin for Bokeh. I've tested it with v3 and it still works great.

sdc50: One other thought, do we want to mention anything about Tethys v5.0?

swainn:
Tutorial language: "...this package may not be included in some Tethys installations, so you should include it as a dependency of your app to ensure it is installed."

docs/tethys_sdk/gizmos/bokeh_view.rst
swainn: "This gizmo requires the `bokeh` library to be installed. Starting with Tethys 5.0 or if you are using `microtethys`, you will need to install `bokeh` using conda or pip as follows:"

swainn: > As a side note I think we can take off the pin for Bokeh. I've tested it with v3 and it still works great.

I get an error in the showcase app when I have Bokeh 3 installed. See: #980

sdc50:
@swainn We need to rebase this branch on master. I'm inclined to squash the commits and then rebase and force push. That would require you (and anyone else that has this branch checked out) to re-set it (or delete and re-checkout). Do you have any changes to push before I do that?

swainn:
> @swainn We need to rebase this branch on master. I'm inclined to squash the commits and then rebase and force push. That would require you (and anyone else that has this branch checked out) to re-set it (or delete and re-checkout). Do you have any changes to push before I do that?

Sounds like a good plan to me. I don't have any changes that I need to merge.

docs/supplementary/optional_features.rst
```diff
@@ -0,0 +1,270 @@
...
+
+With the release of ``microtethys`` and starting with Tethys v5.0 many features of Tethys Platform will become optional and require additional dependencies to be installed. This is done to limit the size of the environment and allow Tethys Portal administrators more flexibility in deciding what features are needed in their deployment. The following list of features are optional and will require the listed dependencies to be installed for that feature to be enabled:
+
+Security Features
+=================
+
+Session Security
+----------------
+
+Session security enables setting timeouts for user sessions.
```
swainn, line 15:
```suggestion
Session security enables setting timeouts for user sessions and automatically logging them out.
```

docs/supplementary/optional_features.rst
```diff
@@ -0,0 +1,270 @@
...
+
+**dependencies**
+ - ``django-session-security``
+
+
+Track Failed Login Attempts
+---------------------------
+
+Tracking failed loging allows Tethys to lock user accounts after a certain number of failed attempts, and alerts users of the number of failed attempts when they login.
```
swainn, line 24:
```suggestion
Tracking failed logins allows Tethys to lock user accounts after a certain number of failed attempts, and alerts users of the number of failed attempts when they login.
```

docs/supplementary/optional_features.rst
```diff
@@ -0,0 +1,270 @@
...
+**dependencies**
+    - ``django-oauth-toolkit``
+
+Portal Enhancements
+===================
+
+Terms and Conditions
+--------------------
+
+Enables portal administrators to define terms and conditions that users must accept to user the portal.
```
swainn, line 113:
```suggestion
Enables portal administrators to define terms and conditions that users must accept to use the portal.
```

docs/supplementary/optional_features.rst
```diff
@@ -0,0 +1,270 @@
...
+**dependencies**
+    - ``condorpy``
+
+External Services
+=================
+
+Dataset Services
+----------------
+
+Enables the :term:`dataset services` API.
```
swainn, line 248:
```suggestion
Enables the :term:`dataset services` APIs for CKAN and GeoServer.
```

swainn (commented):
Aside from my other comments I have one other concern: how are we testing the micro version / optional dependencies. I think we decided that it was too complex to test both ways, but I wonder if there is a straightforward way to modify tests that are affected by an optional dependency. Something like a `try: except:` that catches the exception raised? Maybe as a decorator so we can easily add it to the tests that need it. Have you tried anything like that?

sdc50:
fixes #980 and fixes #985

sdc50:
@swainn I've finished the docs for this PR, so it should be ready to go.

swainn (commented):
I Approve

PR #969, opened 2023-06-29
Addition of Prefix path

romer8:
Adding prefix path to the tethys platform through a new portal_config setting called PREFIX_TO_PATH:

- [X] Updated test
- [X] Redirect empty / to /PREFIX_TO_PATH
- [X] Set the STATIC_URL setting when the PREFIX_TO_PATH is set
- [X] Set the LOGIN_URL setting when the PREFIX_TO_PATH is set to handle redirect of the form /admin


tethys_portal/urls.py
```diff
@@ -38,13 +40,23 @@
...
+prefix_path = ""
+if hasattr(settings, "PREFIX_TO_PATH") and settings.PREFIX_TO_PATH:
+    prefix_path = f"{settings.PREFIX_TO_PATH}"
+
+
 admin.site.login = staff_member_required(
-    admin.site.login, redirect_field_name="", login_url="/accounts/login/"
+    admin.site.login,
+    redirect_field_name="",
+    login_url=f"/{prefix_path}/accounts/login/",
```
swainn, line 51: When `prefix_path` is `""`, does this result in a url like: `"//accounts/login/"`? Does that cause issues? Also, what about `/`s in the `prefix_path` itself?

tethys_portal/urls.py
```diff
@@ -226,3 +238,11 @@
 handler403 = tethys_portal_error.handler_403
 handler404 = tethys_portal_error.handler_404
 handler500 = tethys_portal_error.handler_500
+
+
+if prefix_path is not None and len(prefix_path) != 0:
+    # breakpoint()
```
swainn, line 244: Remove commented out `breakpoint()`

tethys_portal/urls.py
```diff
@@ -38,13 +40,23 @@
...
 admin.site.login = staff_member_required(
-    admin.site.login, redirect_field_name="", login_url="/accounts/login/"
+    admin.site.login,
+    redirect_field_name="",
+    login_url=f"/{prefix_path}/accounts/login/",
 )
 
+
 admin.autodiscover()
 admin.site.login = staff_member_required(
```
swainn, line 56: I know this was here before, but do you know why admin.site.login is being set twice? Does it still work if one of the definitions is removed?

tests/unit_tests/test_tethys_apps/test_base/test_handoff.py
```diff
@@ -3,6 +3,11 @@
 from types import FunctionType
 from unittest import mock
 from tethys_sdk.testing import TethysTestCase
+from django.conf import settings
+
+prefix_to_path = ""
+if settings.PREFIX_TO_PATH is not None and len(settings.PREFIX_TO_PATH) != 0:
+    prefix_to_path = f"/{settings.PREFIX_TO_PATH}"
```
swainn, line 10: Rather than getting settings in tests you should set them using `self.settings` context mananger or the `@override_settings` decorator. This will make the tests indepedendent of whatever settings are currently set in the devs environment. See: https://stackoverflow.com/questions/36611393/how-to-mock-django-settings-attribute-used-in-another-module

swainn (changes requested):
See comments.

tethys_portal/urls.py
romer8: I have taken it away and it seems to work well, and It also pass the tests

tests/unit_tests/test_tethys_apps/test_base/test_handoff.py
romer8: I have a used also another method inside the Test classes to reload the URLs since they get formed with the extra settings.

tethys_portal/urls.py
romer8: removed

tethys_portal/urls.py
romer8: I have used the following to avoid such conflicts:
```python
LOGIN_URL = portal_config_settings.pop("LOGIN_URL", "/accounts/login/")
PREFIX_URL = portal_config_settings.pop("PREFIX_URL", "/")
if PREFIX_URL is not None and PREFIX_URL != "/":
    PREFIX_URL = f"{PREFIX_URL.strip('/')}"
    STATIC_URL = f"/{PREFIX_URL}/{STATIC_URL.strip('/')}/"
    LOGIN_URL = f"/{PREFIX_URL}/{LOGIN_URL.strip('/')}/"
    FIDO_LOGIN_URL = f"/{PREFIX_URL}/{FIDO_LOGIN_URL.strip('/')}/"
````

File: tethys_portal/settings.py [hunk 5]
```diff
@@ -509,3 +511,8 @@
 CORS_CONFIG = portal_config_settings.pop("CORS_CONFIG", {})
 for setting, value in CORS_CONFIG.items():
     setattr(this_module, setting, value)
+
+PREFIX_TO_PATH = portal_config_settings.pop("PREFIX_TO_PATH", "")
+if PREFIX_TO_PATH is not None and len(PREFIX_TO_PATH) != 0:
+    STATIC_URL = f"/{PREFIX_TO_PATH}/static/"
```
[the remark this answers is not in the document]
romer8: I have added the prefix to FIDO_LOGIN_URL, but it also seems to have been outdated

tethys_portal/settings.py
```diff
@@ -509,3 +511,8 @@
...
+    LOGIN_URL = f"/{PREFIX_TO_PATH}/accounts/login/"
```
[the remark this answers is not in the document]
romer8: the following has been used:
```python
LOGIN_URL = portal_config_settings.pop("LOGIN_URL", "/accounts/login/")
PREFIX_URL = portal_config_settings.pop("PREFIX_URL", "/")
if PREFIX_URL is not None and PREFIX_URL != "/":
    PREFIX_URL = f"{PREFIX_URL.strip('/')}"
    STATIC_URL = f"/{PREFIX_URL}/{STATIC_URL.strip('/')}/"
    LOGIN_URL = f"/{PREFIX_URL}/{LOGIN_URL.strip('/')}/"
    FIDO_LOGIN_URL = f"/{PREFIX_URL}/{FIDO_LOGIN_URL.strip('/')}/"
```

tethys_portal/settings.py
```diff
@@ -509,3 +511,8 @@
 CORS_CONFIG = portal_config_settings.pop("CORS_CONFIG", {})
 for setting, value in CORS_CONFIG.items():
     setattr(this_module, setting, value)
+
+PREFIX_TO_PATH = portal_config_settings.pop("PREFIX_TO_PATH", "")
```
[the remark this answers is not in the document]
romer8: PREFIX_URL has been chosen

tethys_portal/urls.py
```diff
@@ -205,7 +197,7 @@
     re_path(r"^user/", include((user_urls, "user"), namespace="user")),
     re_path(r"^apps/", include("tethys_apps.urls")),
     re_path(r"^extensions/", include(extension_urls)),
-    re_path(r"^developer/", include(developer_urls)),
+    re_path(r"^developer/", include((developer_urls))),
```
swainn, line 200: Why the extra set of parenthesis?

tethys_gizmos/gizmo_options/jobs_table.py
```diff
@@ -437,6 +439,11 @@ def __init__(
         if self.show_actions:
             self.num_cols += 1
 
+        # Derive base_ajax_url dynamically for use in JavaScript
+        # e.g.: /developer/gizmos/ajax/9999/action/delete
+        self.base_ajax_url = reverse("gizmos:delete_job", kwargs={"job_id": 9999})
+        self.base_ajax_url = self.base_ajax_url.replace("9999/action/delete", "")
```
sdc50, line 445: This is an interesting way to do this. It feels like a work around. Any thoughts on a more direct way to get a base URL? We may need to do similar things in other places.

tethys_gizmos/static/tethys_gizmos/less/tethys_map_view.less
```diff
@@ -75,7 +75,6 @@
 
   .tethys-map-view-draw-icon {
     display: inline-block;
-    background: url('/static/tethys_gizmos/images/map-view-drawing-icons.gif') no-repeat;
```
sdc50, line 78: I'm not sure what this style is used for, but does removing the background image make sense?

sdc50: i.e. do we need to find another way to get the URL for the background image?

sdc50: Ah, I found it. You've just put this directly in the template. Maybe a comment here would be a helpful clarification?

sdc50 (commented):
I think this does a lot of things in a better way. It's good to get some hard-coded urls removed. I'm curious what the motivation for the `url_prefix` is though. Is someone needing to host multiple sites on a single domain, so they want a better way to filter urls on the proxy server?

swainn:
> I think this does a lot of things in a better way. It's good to get some hard-coded urls removed. I'm curious what the motivation for the `url_prefix` is though. Is someone needing to host multiple sites on a single domain, so they want a better way to filter urls on the proxy server?

Hey @sdc50 thanks for reviewing this. The motivation is as you surmised, we have a need for installing a Tethys Portal and a Django CMS instance on the same domain. Since they are both Django-based, we were getting conflicts with some of the URLs such as the admin site URLs. In general, I think being able to prefix Tethys will allow for flexibility in deploying it in the future.

tethys_gizmos/gizmo_options/jobs_table.py
swainn: It's a workaround only because we don't have a URL that resolves to the pattern we needed for our template URL. Maybe we should add a URL for `/developer/gizmos/ajax/`? Not sure what the view would render as, so it would be a different hacky workaround. Either way, I think using `reverse()` in controllers or `url` in templates is what we should do to get URLs.

tethys_gizmos/static/tethys_gizmos/less/tethys_map_view.less
swainn: We had to move it to the template so that we could retrieve it dynamically, rather than having it hard-coded.

swainn: > Ah, I found it. You've just put this directly in the template. Maybe a comment here would be a helpful clarification?

@romer8 could you add a comment where we removed this line that references the template it is in?

romer8: Perfect, I have added some comments in the palces where the code was removed and added.

sdc50 (approved):

swainn:
@sdc50 thoughts on merging this despite the issue with the macos tests?

swainn (approved):

swainn:
> @sdc50 thoughts on merging this despite the issue with the macos tests?

nvm. I was able to get the mac tests running again by switching the tests to use conda-libmamba-solver instead of using the mamba package.

swainn:
@romer8 I think we just need to get that coverage back up to 100 and we can merge.

swainn:
Rebased and pushed to branch in the tethysplatform/tethys repo to make sure Docker and Conda builds are working. See #974.

PR #988, opened 2023-09-26
List urls

sdc50:
Add `--urls` option to `tethys list` command to get list of all apps' `UrlMaps`.


swainn (approved):
Thanks for the addition! This will be very helpful I think.

PR #989, opened 2023-09-26
Proxy Apps with icon to distinguish from native apps

romer8:
- [x] New field added in the ProxyApps model called display_external
- [x] Conditional rendering in case the display_external of the proxy app is disable
- [x] Migration created
- [x] Documentation updated

Icon appearance:
image

Admin appearance:
image


sdc50:
@romer8 I like this approach. I think it may be more clear to have the checkbox label say `Display External Indicator`. I'm also curious how it looks if the app name is longer. How would it look if the icon were about the same size as the `info` icon that shows up at the top of the app card?

docs/tethys_portal/admin_pages.rst
```diff
@@ -279,6 +279,7 @@ Proxy apps are links to external web applications or websites. They are represen
 * **Show in apps library**: Display the app on the Apps Library page. The app is still accessible by URL, but no app tile will be shown in the apps library.
 * **Open in new tab**: Open the proxy app link in a new tab or window when on. Open it the same tab or window when off.
 * **Order**: Force the order that apps are listed on the Apps Library page. Default order is alphabetical.
+* **External Display**: By default, this feature is turned off. It adds an icon to distinguish proxy apps from regular apps.
```
swainn, line 282:
```suggestion
* **Display External Icon**: By default, this feature is turned off. When enabled, it adds an icon to the app tile to distinguish the proxy app from regular apps, indicating that it is an external link.
```

tethys_apps/models.py
```diff
@@ -1276,6 +1276,7 @@ class ProxyApp(models.Model):
     show_in_apps_library = models.BooleanField(default=True)
     open_in_new_tab = models.BooleanField(default=True)
     order = models.IntegerField(default=0)
+    display_external = models.BooleanField(default=False)
```
swainn, line 1279:
```suggestion
    display_external_icon = models.BooleanField(default=False)
```

swainn (commented):
The icon you provided is the right symbol, but too big and the circle around it makes it look off-center b/c of the shape of the icon. I recommend we use the bootstrap icon with no circle around it: https://icons.getbootstrap.com/icons/box-arrow-up-right/. Make it a little smaller too.

I like the placement of the icon in the bottom right corner of the app tile, but please add white space to offset it from the corners a bit.

tethys_apps/models.py
romer8: This was changed as suggested

docs/tethys_portal/admin_pages.rst
romer8: This was changed as suggested

romer8:
@swainn @sdc50 I have changed the style of the icon to look like this:

image

I have made the name long as well to see if it looks good.
The name has been changed in the model and admin page also to look like **Display External Icon**

swainn:
> I have made the name long as well to see if it looks good. The name has been changed in the model and admin page also to look like **Display External Icon**

Excellent. I like this version a lot more. I'd like to get feedback from @sdc50 and/or @shawncrawley on this too.

swainn (approved):

shawncrawley (approved):
Yeah, I like this. Pretty straightforward and looks good.

PR #1004, opened 2024-01-05
Add Custom Settings to Apps Endpoint of Portal API

swainn:
Adds the custom settings to the object that is returned when calling the `apps` endpoint of the Tethys Portal REST API. This makes the custom settings available to apps with the React frontend, which uses the REST API to get info about the app.

- The custom settings are only added if the `request` user is authenticated. Otherwise, the endpoint returns only the basic metadata about the app.
- This can include all types of Custom Settings: JSON, Secret, and Simple.
- Which custom settings are included is controlled with an new field on the `CustomSettingsBase` model: `include_in_api`. The setting is included in the API response if `include_in_api` is `True`. Defaults to `False`.

Example of object returned with custom settings:

```json
{
  "title": "tRIBS Model Builder",
  "description": "An application for building new tRIBS models for SRP studies.",
  "tags": "",
  "package": "tribs",
  "urlNamespace": "tribs",
  "color": "#115590",
  "icon": "/static/tribs/images/icon.gif",
  "exitUrl": "/apps/",
  "rootUrl": "/apps/tribs/home/home/",
  "settingsUrl": "/admin/tethys_apps/tethysapp/3/change/",
  "customSettings": {
    "cesium_api_key": {
      "type": "STRING",
      "value": "some-api-key"
    },
    "json_setting": {
      "type": "JSON",
      "value": {"foo": "bar"}
    },
    "secret_setting": {
      "type": "SECRET",
      "value": "some-secret"
    }
  }
}
```

Also in this PR:
* Unpinned sphinx version - seems to be working again
* Updated configuration for newer version of sphinx theme


shawncrawley (approved):
Sorry it took a bit longer to get to this, but I've looked it over and approve!

PR #1013, opened 2024-03-07
Microtethys image

romer8:
- [x] Create docker image for Micro Tethys
- [x] Allow salt tethyscore to use a sqlite or postgresql db depending on the TETHYS_DB_ENGINE variable
- [x] Allow run.sh to recognize between sqlite and postgresql


docker/run.sh
```diff
@@ -44,32 +45,40 @@ done
...
+      fi
+      >&2 echo "DB is unavailable - sleeping"
+      db_check_count=`expr $db_check_count + 1`
+      sleep 5
+    done
+  elif [[ $db_engine == "django.db.backends.sqlite3" ]]; then
+    echo_status "Using SQLite3 as the database"
+    # Add any SQLite-specific setup here if necessary
+  else
+    echo_status "Unsupported database engine: $db_engine"
```
sdc50, line 78: I think this is fine for now, but Tethys will support other engines. I use the Tethys docker with a pre-configured SQL Server DB.

docker/salt/tethyscore.sls
```diff
@@ -101,6 +93,20 @@ Generate_Tethys_Settings_TethysCore:
         {{ OTHER_SETTINGS }}
     - unless: /bin/bash -c "[ -f "{{ TETHYS_PERSIST }}/setup_complete" ];"
 
+{% if TETHYS_DB_ENGINE == 'django.db.backends.postgresql' %}
```
sdc50, line 96: I don't think these should be conditional. Even if using SQLite a user may want to set the `TETHYS_DB_NAME`. I needed to add the following to support SQL Server. 

```jinja
        {%- if TETHYS_DB_OPTIONS %}
        --set DATABASES.default.OPTIONS {{ TETHYS_DB_OPTIONS }}
        {%- endif %}
```

I guess I'm fine keeping this as is for now and saying that we only support SQLite and PostgreSQL. I can make modifications later to support other Databases, web-servers, and base images.

docker/salt/tethyscore.sls
```diff
@@ -149,7 +156,8 @@ Create_Database_User_and_SuperUser_TethysCore:
         -N "{{ TETHYS_DB_SUPERUSER }}"
         -P "{{ TETHYS_DB_SUPERUSER_PASS }}"
     - shell: /bin/bash
-    - unless: /bin/bash -c "[ -f "{{ TETHYS_PERSIST }}/setup_complete" ];"
+    - unless: /bin/bash -c "[ -f '{{ TETHYS_PERSIST }}/setup_complete' ];"
```
sdc50, line 159: There is the case of wanting to use the Tethys Docker to connect to an existing PostgreSQL database, in which case the `TETHYS_DB_ENGINE` would be `'django.db.backends.postgresql'` but you would not want to create these databases. I've taken the approach of adding a separate variable `SKIP_DB_SETUP` to be able to control this independently of the `TETHYS_DB_ENGINE` value. 

```salt
    - unless: /bin/bash -c "[ -f "{{ TETHYS_PERSIST }}/setup_complete" ] || [ "$SKIP_DB_SETUP" = true ];"
```

sdc50 (changes requested):
It looks like this branch needs to be rebased onto `main`.  That should probably fix the tests. I've left several comments, but would be fine holding off on most of them for now. I think we should ensure that this will work with an already configured Postgres DB.

docker/run.sh
romer8: The comment regarding other databases not being supported has been removed.

docker/salt/tethyscore.sls
romer8: added conditional for all the different db settings when setting up all the portal config settings. In addition, added, the SKIP_DB_SETUP env variable: 
fcd5a9140214a0d0978c6a1a2a7aee8eee1773f6

docker/salt/tethyscore.sls
romer8: Added conditional for all the different db settings when setting up all the portal config settings:

https://github.com/tethysplatform/tethys/commit/fcd5a9140214a0d0978c6a1a2a7aee8eee1773f6

romer8:
[#docs/installation/production/docker/tethys_docker_reference.rst](https://github.com/tethysplatform/tethys/pull/1013/commits/2d5e04f21f80b59421e7105507bb16716aa5b4cd)
I have also added extra documentation for the `SKIP_DB_SETUP` and `MICRO_TETHYS` variables.

docker/salt/tethyscore.sls
```diff
@@ -14,6 +14,11 @@
 {% set POSTGRES_PASSWORD = salt['environ.get']('POSTGRES_PASSWORD') %}
 {% set TETHYS_DB_PORT = salt['environ.get']('TETHYS_DB_PORT') %}
 {% set TETHYS_DB_USERNAME = salt['environ.get']('TETHYS_DB_USERNAME') %}
+{% if salt['environ.get']('SKIP_DB_SETUP') %}
```
sdc50, line 17: Rather than having this logic here, lets just add the `SKIP_DB_SETUP` environment variable to the `Dockerfile`

docker/run.sh
```diff
@@ -44,32 +45,37 @@ done
...
-echo "postgres.pass: '${TETHYS_DB_PASSWORD}'" >> /etc/salt/minion
-echo "postgres.bins_dir: '${CONDA_HOME}/envs/${CONDA_ENV_NAME}/bin'" >> /etc/salt/minion
 
 if [[ $test = false ]]; then
   # Set extra ENVs
   export NGINX_USER=$(grep 'user .*;' /etc/nginx/nginx.conf | awk '{print $2}' | awk -F';' '{print $1}')
 
   # Apply States
   echo_status "Checking if DB is ready"
+  if [[ $db_engine == "django.db.backends.postgresql" ]]; then
```
sdc50, line 55: I think we also want this whole `if` block wrapped in another `if` block that checks if `SKIP_DB_SETUP` is `False`.

sdc50:
@romer8 I've added a few more suggestions to your recent changes. I think this is almost there!

docker/run.sh
romer8: I have added the extra if statement to handle the skip db env variable. That was good catch.
https://github.com/tethysplatform/tethys/pull/1013/commits/8ce70dd5ce57bdc488fd42ae297dfd1824476614

docker/salt/tethyscore.sls
romer8: removed the checking in the tethyscore.sls and added a new env variable into the Dockerfile
https://github.com/tethysplatform/tethys/pull/1013/commits/d99c5cb08cafdd4fba79414880a7c87fa3b73866

sdc50 (approved):

swainn (commented):
See comments

docker/run.sh
```diff
@@ -44,32 +46,43 @@ done
...
+        sleep 5
+      done
+    
+    else
+      # Database setup should be skipped
+      echo "Skipping database setup: check SKIP_DB_SETUP environment variable."
+    fi
+  else
+    echo_status "Using SQLite3 as the database"
+  fi
```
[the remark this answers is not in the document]
romer8: I agree. It looks better in the suggested way. I corrected it:

docker/run.sh
```diff
@@ -44,32 +46,43 @@ done
...
-echo "postgres.bins_dir: '${CONDA_HOME}/envs/${CONDA_ENV_NAME}/bin'" >> /etc/salt/minion
 
 if [[ $test = false ]]; then
   # Set extra ENVs
   export NGINX_USER=$(grep 'user .*;' /etc/nginx/nginx.conf | awk '{print $2}' | awk -F';' '{print $1}')
 
   # Apply States
   echo_status "Checking if DB is ready"
+  if [[ $db_engine == "django.db.backends.postgresql" ]]; then
+    if [[ "$skip_db_setup" != "true" ]]; then
```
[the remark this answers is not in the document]
romer8: now the variable `db_engine` is a boolean:
https://github.com/tethysplatform/tethys/pull/1013/commits/0fc3464f0efbc3746361757fd90601edcb592400

docs/installation/production/docker/tethys_docker_reference.rst
```diff
@@ -98,7 +98,8 @@ Use these build arguments with the ``build`` command to customize how the image
 +===========================+==========================================================================================+
 | PYTHON_VERSION            | The version of Python to build the Tethys environment with. Defaults to "3.*".           |
 +---------------------------+------------------------------------------------------------------------------------------+
-
+| MICRO_TETHYS              | Set to True if you intent to use `micro-tethys`. Defaults to False.                      |
```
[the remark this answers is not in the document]
romer8: Fixed:
https://github.com/tethysplatform/tethys/pull/1013/commits/2ed1b30f801d10326279811b1e60d23db19e5c8b

swainn (approved):

PR #1015, opened 2024-03-08
Standalone App Mode

ckrew:
This PR allows a tethys portal to be configured to act as a single app instead of a host for multiple apps.

In order to enable this mode, users will need to update their portal_config.yml under `TETHYS_PORTAL_CONFIG` then `MULTIPLE_APP_MODE`. The value is a boolean. The default value for this is True and by changing it to False, the standalone app mode will be enabled. An optional setting is `STANDALONE_APP`. This can be used to determine which app will be the standalone app. If this is not provided, then the code will just use the first application in the DB. 

standalone_app_config

The configured standalone application will become the home page for the tethys portal. The app library page is redirected to the standalone app as well. You can see the example below where I am accessing the application from the root instead of apps.

image

You can additionally see in the screenshot above, that the user menu is now available within the application itself. I have updated the headers in the app base html to show the user menu button if the Standalone App mode is turned on. A user can use that button to check portal/user settings just like before.

When a user now checks the other available tethys endpoints like the user profiles and admin settings, the headers are updated to reflect the Standalone App mode. As shown in the screenshot below, the "Tethys Portal" button on the left can still be updated to reflect a brand or company.  The "Apps" button on the right will now display the app package name and redirect to the application.

image


tests/unit_tests/test_tethys_apps/test_cli/test_install_commands.py
```diff
@@ -3,7 +3,7 @@
 from django.core.exceptions import ValidationError
 from unittest import mock
 from conda.cli.python_api import Commands
-from tethys_apps.cli import install_commands
+from tethys_cli import install_commands
```
sdc50, line 6: Wait, what?! How was this even working before? I think we just need to delete this file. It looks like it was replaced by `tests/unit_tests/test_tethys_cli/test_install_commands.py`.

tethys_apps/templates/tethys_apps/app_base.html
```diff
@@ -192,16 +211,54 @@
...
+                    <div class="header-button settings-button">
+                      <a href="javascript:void(0);" onclick="TETHYS_APP_BASE.exit_app('{% url 'admin:index' %}tethys_apps/tethysapp/{{ tethys_app.id }}/change/');" data-bs-toggle="tooltip" data-bs-placement="bottom" title="Settings"><i class="bi bi-gear"></i></a>
+                    </div>
+                    {%  endif %}
                   {% endblock %}
                   {% block exit_button_override %}
-                    <div class="header-button exit-button">
-                      <a href="javascript:void(0);" onclick="TETHYS_APP_BASE.exit_app('{% url 'app_library' %}');"data-bs-toggle="tooltip" data-bs-placement="bottom" title="Exit"><i class="bi bi-x"></i></a>
-                    </div>
+                    {% if not single_app_mode %}
```
sdc50, line 221: This looks like it's mostly duplicated from `tethys_portal/header.html`. Is there a way we could extract this into its own template (e.g. `user_menu.html`) and then just include it in both places?

ckrew:
@sdc50 and @swainn The two suggested changes have been implemented

sdc50 (approved):
Looks great! Thanks @ckrew!

sdc50:
@swainn I think this is good to be merged if you approve now.

swainn (approved):

Issue #998, opened 2023-11-16
Typo in Map Layout Basemaps  [BUG]


AbbyKloss:
**Describe the bug**
A typo in the Map Layout class's basemaps attribute leads to that layer and any proceeding layer to not load.

**To Reproduce**
1. (optional) Install the docker container through the [Docker Compose file](https://github.com/tethysplatform/tethys/blob/app-developer-docker-dev-image/docker/tethys-dev/docker-compose.geoserver.yml)
2. Clone https://github.com/tethysplatform/tethysapp-map_layout_tutorial/tree/configure-data-plotting-solution
3. Install the app
4. in the controllers.py, on line 22, add the following: `basemaps = ["OpenStreetMap", "ERSI"]`
Note: the correct line to add for a working page would be `basemaps = ["OpenStreetMap", "ESRI"]`
5. In the app home page, the only layer that will load is the OpenStreetMap layer

**Expected behavior**
What I would expect is for the misspelled error to be skipped over, rather than taking down every proceeding layer with it.

**Errors/Traceback**

JavaScript Console Output:
```
jQuery.Deferred exception: m_wms_feature_selection_changed_callbacks is undefined onSelectionChange@http://localhost:8000/static/tethys_gizmos/js/tethys_map_view.js:2570:7
init_feature_selection@http://localhost:8000/static/tethys_layouts/map_layout/map_layout.js:1579:23
setup_map@http://localhost:8000/static/tethys_layouts/map_layout/map_layout.js:212:6
@http://localhost:8000/static/tethys_layouts/map_layout/map_layout.js:1947:6
mightThrow@https://cdn.jsdelivr.net/npm/jquery@3.6.0/dist/jquery.js:3766:29
Deferred/then/resolve/</process<@https://cdn.jsdelivr.net/npm/jquery@3.6.0/dist/jquery.js:3834:12
setTimeout handler*Deferred/then/resolve/<@https://cdn.jsdelivr.net/npm/jquery@3.6.0/dist/jquery.js:3872:16
fire@https://cdn.jsdelivr.net/npm/jquery@3.6.0/dist/jquery.js:3500:31
fireWith@https://cdn.jsdelivr.net/npm/jquery@3.6.0/dist/jquery.js:3630:7
fire@https://cdn.jsdelivr.net/npm/jquery@3.6.0/dist/jquery.js:3638:10
fire@https://cdn.jsdelivr.net/npm/jquery@3.6.0/dist/jquery.js:3500:31
[...]

PR #1017, opened 2024-03-13
Fix Issue with Typo in MapLayout Basemaps

gagelarsen:
This PR is to resolve issue https://github.com/tethysplatform/tethys/issues/998.

The `addLayer` call for adding a map layer was outside of the check for supported basemap type. I just moved the call into the correct scope.


sdc50 (approved):

shawncrawley (approved):
Beautiful

Issue #1010, opened 2024-02-12
[FEATURE] Local logo URL for proxy app icon


ckrew:
**Is your feature request related to a problem? Please describe.**
Proxy app allows for logo urls to be http endpoints but not local pngs.

**Describe the solution you'd like**
It would be nice if users could specify local pngs for proxy apps, like they can do for tethys apps.

ckrew:
@swainn Looking into this and I wonder if the main reason this wasn't implemented initially was because the proxyapp metadata is all contain within the database without any file footprint. In other words, there is no app workspace or any file location to store a png like there is for tethysapps. 

Any thoughts on how to implement? If there is a general tethys workspace, we could create a proxy app folder there. We would also read the image file and store it in the DB but that might not be ideal for size reasons.

swainn:
@ckrew that is a good point. One scenario where this may be applicable would be when a Tethys extension is used to install a Theme for a portal. The Tethys extension has it's own public/static directory where images and other static files can be stored. This could be used as a location for custom proxy app images.

Of course, in a production installation, a portal admin could add images to any of the existing static files directories after they have been collected for this purpose. This is not ideal, but manageable, especially if automated in a container. We've also talked about adding a way to upload images for use in cases like these...

PR #1021, opened 2024-03-19
Proxy App Icon Update

ckrew:
This PR updates the proxy app logo URL to be a proxy app icon string. This will change the proxy app icon behavior to be exactly like a normal tethys app. Users can use a http url string as before but can also now reference a local static file for a proxy app icon/logo. This PR will address issue https://github.com/tethysplatform/tethys/issues/1010.

Changes include:

- Updating the proxy app model field from `logo_url (URLField)` to `icon (CharField)`
- Updating html template to handle tethys app and proxy app icons the same way, with an if statement for proxied and external icons proxy apps
- Updating tests and files to use `icon` instead of `logo_url` for the proxy app model field


tethys_apps/templates/tethys_apps/app_card.html
```diff
@@ -14,18 +14,13 @@
     {% endif %}
         <div class="card app-card text-center">
             <div class="card-body">
-            {% if app.proxied %}
-                <img class="card-img-top" {% if app.icon %}src="{{ app.icon }}"{% else %}src="{% static 'tethys_apps/images/default_app_icon.gif' %}"{% endif %} />
-                {% if app.display_external_icon %}
+                <img class="card-img-top" {% if app.icon %}{% if 'http' in app.icon %}src="{{ app.icon }}"{% else %}src="{% static app.icon %}"{% endif %}{% else %}src="{% static 'tethys_apps/images/default_app_icon.gif' %}"{% endif %} alt="">
+                {% if app.display_external_icon and app.proxied %}
```
swainn, line 18:
```suggestion
                {% if app.display_external_icon %}
```

tethys_apps/migrations/0001_initial_41.py
```diff
@@ -28,7 +28,7 @@ class Migration(migrations.Migration):
                 ),
                 ("name", models.CharField(max_length=100, unique=True)),
                 ("endpoint", models.URLField(max_length=512)),
-                ("logo_url", models.URLField(blank=True, max_length=512)),
+                ("icon", models.URLField(blank=True, max_length=512)),
```
swainn, line 31: Don't modify by hand. Run `tethys manage makemigrations tethys_apps`

swainn: To apply the migration `tethys manage migrate`

swainn: Also use migrations.RunPython steps to migrate the data too. Forward and back. See this for an example: https://github.com/tethysplatform/tethys/blob/main/tethys_config/migrations/0001_initial_40.py

swainn (changes requested):
You know what to do. ;D

ckrew:
@c-krew Fix migration for the new char field as well. May need to look at saving logo url data and then reloading it into the new icon field after the migration is done.

tethys_apps/migrations/0001_initial_41.py
ckrew: The new migration file has been added and tested

ckrew:
@sdc50 and @swainn The changes from the tethys scrum have been implemented

tethys_apps/migrations/0006_auto_20240401_2032.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def backward(apps, schema_editor):
+    """From proxyapp with icon field to logo_url field"""
+    ProxyApp = apps.get_model("tethys_apps", "proxyapp")
+
+    db_alias = schema_editor.connection.alias
+
+    for cs in ProxyApp.objects.using(db_alias).all():
+        cs.logo_url = cs.icon
```
sdc50, line 24: This could fail if a user migrates, adds static files as icons and then un-migrates. I'm not too concerned about that scenario though. I'm not sure what would happen (if it would drop the data or refuse to undo the migration).

sdc50 (approved):

tethys_apps/migrations/0006_auto_20240401_2032.py
ckrew: oh good point. I can do some testing with this. What would be the desired logic? Raise an exception and let the user know that the logo_url needs to be a url? Raise a warning and make the field empty or make it a fake http url with the static file address?

sdc50: @ckrew I'm not too concerned about this, so I think we should just move forward as is. @swainn thoughts?

swainn: I think the only ones who would unmigrate, would be us during development. It would be very unlikely for our users, so I'm not worried about it either.

swainn (approved):

PR #1022, opened 2024-03-20
bokehjsdir() has been deprecated in favor of bokehjs_path()

gronka:
bokehjsdir() has been deprecated in favor of bokehjs_path()

Using bokehjsdir() now causes an error when running docker build

https://github.com/bokeh/bokeh/blob/69c084e692d79632c97ea1cfc5ceb10d0217e6f4/src/bokeh/settings.py#L779

### DIFF
--- a/tethys_portal/settings.py
+++ b/tethys_portal/settings.py
@@ -37,8 +37,8 @@ from tethys_cli.gen_commands import generate_secret_key
 from tethys_portal.optional_dependencies import optional_import, has_module
 
 # optional imports
-bokeh_settings, bokehjsdir = optional_import(
-    ("settings", "bokehjsdir"), from_module="bokeh.settings"
+bokeh_settings, bokehjs_path = optional_import(
+    ("settings", "bokehjs_path"), from_module="bokeh.settings"
 )
 bokeh_django = optional_import("bokeh_django")
 
@@ -407,8 +407,8 @@ STATIC_URL = portal_config_settings.pop("STATIC_URL", "/static/")
 STATICFILES_DIRS = [
     BASE_DIR / "static",
 ]
-if has_module(bokehjsdir):
-    STATICFILES_DIRS.append(bokehjsdir())
+if has_module(bokehjs_path):
+    STATICFILES_DIRS.append(bokehjs_path())
 
 STATICFILES_USE_NPM = TETHYS_PORTAL_CONFIG.pop("STATICFILES_USE_NPM", False)
 if STATICFILES_USE_NPM:
--- a/tethys_portal/settings.py
+++ b/tethys_portal/settings.py
@@ -37,9 +37,7 @@ from tethys_cli.gen_commands import generate_secret_key
 from tethys_portal.optional_dependencies import optional_import, has_module
 
 # optional imports
-bokeh_settings, bokehjs_path = optional_import(
-    ("settings", "bokehjs_path"), from_module="bokeh.settings"
-)
+bokeh_settings = optional_import("settings", from_module="bokeh.settings")
 bokeh_django = optional_import("bokeh_django")
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


tethys_portal/settings.py
```diff
@@ -37,8 +37,8 @@
 from tethys_portal.optional_dependencies import optional_import, has_module
 
 # optional imports
-bokeh_settings, bokehjsdir = optional_import(
-    ("settings", "bokehjsdir"), from_module="bokeh.settings"
+bokeh_settings, bokehjs_path = optional_import(
+    ("settings", "bokehjs_path"), from_module="bokeh.settings"
 )
```
sdc50, line 42:
```suggestion
bokeh_settings = optional_import("settings", from_module="bokeh.settings")
```

sdc50: One of the reasons this is breaking now is because we were originally importing `bokehjsdir` incorrectly. It is imported into the `settings` module, but not indented to be imported from there. Rather, a method with the same name is defined on the `settings` object defined in the `settings` module. What we should have been doing is just importing ` bokeh_settings` and then calling `bokehjsdir` like this:

```python
bokeh_settings.bokehjsdir()
```
If we had done it that way it would still work and show the deprecation warning instead of raising an import error.

All of this applies to the new method `bokehjs_path`. It should be called on `bokeh_settings` and not imported directly.

tethys_portal/settings.py
```diff
@@ -407,8 +407,8 @@
 STATICFILES_DIRS = [
     BASE_DIR / "static",
 ]
-if has_module(bokehjsdir):
-    STATICFILES_DIRS.append(bokehjsdir())
+if has_module(bokehjs_path):
+    STATICFILES_DIRS.append(bokehjs_path())
```
sdc50, line 411:
```suggestion
if has_module(bokeh_settings):
    try:
        bokeh_js_dir = bokeh_settings.bokehjs_path()
    except AttributeError:
        # support bokeh versions < 3.4
        bokeh_js_dir = bokeh_settings.bokehjsdir()
    STATICFILES_DIRS.append(bokeh_js_dir)
```

sdc50: This suggested change not only calls `bokehjs_path` on the `bokeh_settings` object as is the intended usage (see comment above), but it also maintains backward compatibility so it will still work with Bokeh versions < 3.4.

sdc50 (changes requested):
I was just about to file an issue for this and submit a PR, but @gronka, you beat me to the punch. See my suggested edits. Also, to keep test coverage we'll have to add a test that triggers the `AttributeError` when calling `bokehjs_path`.

sdc50:
@gronka Add the following test to `tests/unit_tests/test_tethys_portal/test_settings.py`:

```python

    @mock.patch("tethys_portal.optional_dependencies.optional_import")
    def test_bokehjsdir_compatibility(self, mock_oi):
        mock_bokeh_settings = mock.MagicMock()
        mock_oi.return_value = mock_bokeh_settings
        mock_bokeh_settings.bokehjs_path.side_effect = AttributeError()
        reload(settings)
        mock_bokeh_settings.bokehjs_path.assert_called_once()
        mock_bokeh_settings.bokehjsdir.assert_called_once()
```

That should provide complete coverage.

sdc50:
Once we get this merged, we should also cherry pick the commit to the `release` branch as a hotfix, so current builds won't break.

@araglu once the hotfix is applied then docker builds using the latest release of Tethys should work again.

gronka:
@sdc50 Thanks. I added the unit test. The new test passed when I ran the following:

```
kubectl exec --stdin --tty {POD_NAME} -- /bin/bash
pip install bokeh_django
tethys test -c

... testing output ...
TOTAL                                                             10537   7231    31%

37 files skipped due to complete coverage.
```

Just so you know, in case this isn't the correct way to run the tests.

sdc50 (approved):

swainn (approved):

PR #1024, opened 2024-03-21
Additional Settings Files

sdc50:
Add the ability to configure additional settings files that will get dynamically loaded into the `tethys_portal.settings.py` file.


swainn (approved):
Approve once docs are added.

PR #1025, opened 2024-03-22
Bokeh Compatibility

sdc50:
* bokehjsdir() has been deprecated in favor of bokehjs_path()


sdc50:
@sdc50, @shawncrawley, Since this is on a release branch it won't get the changes that fixed `Black` that are on the main branch. I think we should just merge this hotfix as is and not worry about `Black`-ing release branches.

swainn:
> @sdc50, @shawncrawley, Since this is on a release branch it won't get the changes that fixed `Black` that are on the main branch. I think we should just merge this hotfix as is and not worry about `Black`-ing release branches.

I concur

swainn (approved):

PR #1030, opened 2024-04-05
Custom back url for Native Tethys apps

romer8:
- [x] New model field called `back_url` in `TethysApp`
- [x] Added to context processors.
- [x] Added conditional to have back url in `app_base.html` . If it is empty in the model, then use the {% url 'app_library' %}, but if there is a custom value use {{ tethys_app.back_url }}
- [x] Edit current tests


shawncrawley (approved):
Looks good.

swainn (approved):

swainn (approved):

PR #1036, opened 2024-04-14
ARG build variables: django, channels, daphne

romer8:
- [x] ARG build variables to support versions Django, django channels, and daphne according to user needs
- [x] enviroment.yml and micro_environment.yml are edited in line according to the ARG build  env variables

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -143,17 +143,13 @@ WORKDIR ${TETHYS_HOME}/tethys
 
 # Set the versions of Django, Channels, and Daphne if provided in environment.tyml and micro_environment.yml
 RUN if [ -n "$DJANGO_VERSION" ]; then \
-  sed -i "s/- django>=3.2,<6/- django==${DJANGO_VERSION}/" environment.yml micro_environment.yml; \
+  sed -i "s/\s*- django[^-].*/  - django==${DJANGO_VERSION}/" environment.yml micro_environment.yml; \
   fi && \
   if [ -n "$DJANGO_CHANNELS_VERSION" ]; then \
-  sed -i "s/- channels/- channels==${DJANGO_CHANNELS_VERSION}/" environment.yml micro_environment.yml; \
-  else \
-  sed -i "s/- channels/- channels/" environment.yml micro_environment.yml; \
+  sed -i "s/\s*- channels.*/  - channels==${DJANGO_CHANNELS_VERSION}/" environment.yml micro_environment.yml; \
   fi && \
   if [ -n "$DAPHNE_VERSION" ]; then \
-  sed -i "s/- daphne/- daphne==${DAPHNE_VERSION}/" environment.yml micro_environment.yml; \
-  else \
-  sed -i "s/- daphne/- daphne/" environment.yml micro_environment.yml; \
+  sed -i "s/\s*- daphne.*/  - daphne==${DAPHNE_VERSION}/" environment.yml micro_environment.yml; \
   fi
 
 # Create the conda environment based on the environment.yml or micro_environment.yml file
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,9 @@ FROM mambaorg/micromamba:bullseye
 ###################
 ARG PYTHON_VERSION=3.*
 ARG MICRO_TETHYS=false
+ARG DJANGO_VERSION=4.2.*
+ARG DJANGO_CHANNELS_VERSION
+ARG DAPHNE_VERSION
 
 ###############
 # ENVIRONMENT #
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -136,6 +139,22 @@ COPY --chown=$MAMBA_USER:$MAMBA_USER micro_environment.yml ${TETHYS_HOME}/tethys
 
 WORKDIR ${TETHYS_HOME}/tethys
 
+# Set the versions of Django, Channels, and Daphne if provided in environment.tyml and micro_environment.yml
+RUN if [ -n "$DJANGO_VERSION" ]; then \
+  sed -i "s/- django>=3.2,<6/- django==${DJANGO_VERSION}/" environment.yml micro_environment.yml; \
```
swainn, line 144: Can the version part of the django be found using a regular expression? Try this to see if it works:
```suggestion
  sed -i "s/- django.+/- django==${DJANGO_VERSION}/" environment.yml micro_environment.yml; \
```

Dockerfile
```diff
@@ -136,6 +139,22 @@ COPY --chown=$MAMBA_USER:$MAMBA_USER micro_environment.yml ${TETHYS_HOME}/tethys
...
+  fi && \
+  if [ -n "$DJANGO_CHANNELS_VERSION" ]; then \
+  sed -i "s/- channels/- channels==${DJANGO_CHANNELS_VERSION}/" environment.yml micro_environment.yml; \
+  else \
+  sed -i "s/- channels/- channels/" environment.yml micro_environment.yml; \
```
swainn, line 149: I don't think this else is necessary is it?

Dockerfile
```diff
@@ -136,6 +139,22 @@ COPY --chown=$MAMBA_USER:$MAMBA_USER micro_environment.yml ${TETHYS_HOME}/tethys
...
+  fi && \
+  if [ -n "$DAPHNE_VERSION" ]; then \
+  sed -i "s/- daphne/- daphne==${DAPHNE_VERSION}/" environment.yml micro_environment.yml; \
+  else \
+  sed -i "s/- daphne/- daphne/" environment.yml micro_environment.yml; \
```
swainn, line 154: The else probably isn't necessary here either.

Dockerfile
```diff
@@ -136,6 +139,22 @@ COPY --chown=$MAMBA_USER:$MAMBA_USER micro_environment.yml ${TETHYS_HOME}/tethys
...
+RUN if [ -n "$DJANGO_VERSION" ]; then \
+  sed -i "s/- django>=3.2,<6/- django==${DJANGO_VERSION}/" environment.yml micro_environment.yml; \
+  fi && \
+  if [ -n "$DJANGO_CHANNELS_VERSION" ]; then \
+  sed -i "s/- channels/- channels==${DJANGO_CHANNELS_VERSION}/" environment.yml micro_environment.yml; \
+  else \
+  sed -i "s/- channels/- channels/" environment.yml micro_environment.yml; \
+  fi && \
+  if [ -n "$DAPHNE_VERSION" ]; then \
+  sed -i "s/- daphne/- daphne==${DAPHNE_VERSION}/" environment.yml micro_environment.yml; \
```
swainn, line 152: Should probably have a regular expression on this too:
```suggestion
  sed -i "s/- daphne.+/- daphne==${DAPHNE_VERSION}/" environment.yml micro_environment.yml; \
```

Dockerfile
```diff
@@ -136,6 +139,22 @@ COPY --chown=$MAMBA_USER:$MAMBA_USER micro_environment.yml ${TETHYS_HOME}/tethys
 
 WORKDIR ${TETHYS_HOME}/tethys
 
+# Set the versions of Django, Channels, and Daphne if provided in environment.tyml and micro_environment.yml
+RUN if [ -n "$DJANGO_VERSION" ]; then \
+  sed -i "s/- django>=3.2,<6/- django==${DJANGO_VERSION}/" environment.yml micro_environment.yml; \
+  fi && \
+  if [ -n "$DJANGO_CHANNELS_VERSION" ]; then \
+  sed -i "s/- channels/- channels==${DJANGO_CHANNELS_VERSION}/" environment.yml micro_environment.yml; \
```
swainn, line 147: Should probably have a regular expression on this too:
```suggestion
  sed -i "s/- channels.+/- channels==${DJANGO_CHANNELS_VERSION}/" environment.yml micro_environment.yml; \
```

Dockerfile
```diff
@@ -4,6 +4,9 @@ FROM mambaorg/micromamba:bullseye
 ###################
 ARG PYTHON_VERSION=3.*
 ARG MICRO_TETHYS=false
+ARG DJANGO_VERSION=4.2.*
+ARG DJANGO_CHANNELS_VERSION
+ARG DAPHNE_VERSION
```
swainn, line 9: Can args be declared without a value?

Dockerfile
romer8: The following was changed to get anything after `- channels` and replaced if arg build was set
`sed -i "s/\s*- channels.*/  - channels==${DJANGO_CHANNELS_VERSION}/" environment.yml micro_environment.yml;`

Dockerfile
romer8: The following was changed to get anything after - channels and replaced if arg build was set
`sed -i "s/\s*- daphne.*/  - daphne==${DAPHNE_VERSION}/" environment.yml micro_environment.yml; `

Dockerfile
romer8: No, it is not necessary, and it was removed

Dockerfile
romer8: Yes, they can be. They actually help with the if statement, if it is not defined then it will not enter the if statement

Dockerfile
romer8: No, it is not necessary, and it was removed

Dockerfile
romer8: Yes, it can be actually with the code below now the sed command find `- django`, and it replaces it with the `DJANGO_VERSION` arg variable, but it also ignores anything ` -django-` becuase there is some dependencies that contain `django` in the name followed by a `-`:
`sed -i "s/\s*- django[^-].*/  - django==${DJANGO_VERSION}/" environment.yml micro_environment.yml;`

swainn (approved):

sdc50 (approved):

PR #1038, opened 2024-04-17
Fixed and pulled healthcheck into docker/liveness-probe.sh

gronka:
Continuing this discussion:
https://github.com/tethysplatform/tethys/discussions/1018

3 goals here:

1. Kubernetes doesn't apply HEALTHCHECK from a Dockerfile, so I pulled the logic into a script called liveness-probe.sh which is copied into the container, and then can be implemented directly by any container manager (docker, docker-compose, kubernetes, etc).

2. I don't get into the weeds of bash nor docker HEALTHCHECK often. I think this implementation is a bit easier to read and maintain.

3. If I'm not making a mistake, the current implementation of HEALTHCHECK as executed by docker-compose fails to a syntax error:
```
            "Health": {
                "Status": "unhealthy",
                "FailingStreak": 5,
                "Log": [
                    {
                        "Start": "2024-04-17T16:17:10.194952618-05:00",
                        "End": "2024-04-17T16:17:10.236757633-05:00",
                        "ExitCode": 2,
                        "Output": "/bin/sh: 1: Syntax error: \"(\" unexpected\n"
                    },
                    {
--
            "Healthcheck": {
                "Test": [
                    "CMD-SHELL",
                    "function check_process_is_running(){ if [ \"$(ps $1 | wc -l)\" -ne 2 ]; then echo The $2 process \\($1\\) is  not running. 1>&2; return 1; fi };   check_process_is_running $(cat $(grep 'pidfile=.*' /etc/supervisor/supervisord.conf | awk -F'=' '{print $2}' | awk '{print $1}')) supervisor;   check_process_is_running $(cat $(grep 'pid .*;' /etc/nginx/nginx.conf | awk '{print $2}' | awk -F';' '{print $1}')) nginx;   check_process_is_running $(ls -l /run/tethys_asgi0.sock.lock | awk -F'-> ' '{print $2}') asgi;"
                ],
                "StartPeriod": 240000000000
            },
            "Image": "tethysplatform/tethys-core",
```

### DIFF
--- a/docker/liveness-probe.sh
+++ b/docker/liveness-probe.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+check_process_is_running() { 
+	if [ "$(ps $1 | wc -l)" -ne 2 ]; then 
+		echo The $2 process \($1\) is  not running. 1>&2; 
+	else
+		echo "ok"
+	fi 
+}
+
+ret1=$(check_process_is_running $(cat $(grep 'pidfile=.*' /etc/supervisor/supervisord.conf | awk -F'=' '{print $2}' | awk '{print $1}')) supervisor)
+ret2=$(check_process_is_running $(cat $(grep 'pid .*;' /etc/nginx/nginx.conf | awk '{print $2}' | awk -F';' '{print $1}')) nginx)
+ret3=$(check_process_is_running $(ls -l /run/tethys_asgi0.sock.lock | awk -F'-> ' '{print $2}') asgi)
+
+if [ "$ret1" != "ok" ] || [ "$ret2" != "ok" ] || [ "$ret3" != "ok" ]; then
+	echo "liveness-probe.sh: at least 1 test failed"
+	exit 1
+else
+	echo "liveness-probe.sh: all tests passed"
+	exit 0
+fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -207,6 +207,7 @@ EXPOSE 80
 ###############*
 ADD docker/salt/ /srv/salt/
 ADD docker/run.sh ${TETHYS_HOME}/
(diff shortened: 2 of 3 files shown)


docker/liveness-probe.sh
```diff
@@ -0,0 +1,21 @@
...
+	else
+		echo "ok"
+	fi 
+}
+
+ret1=$(check_process_is_running $(cat $(grep 'pidfile=.*' /etc/supervisor/supervisord.conf | awk -F'=' '{print $2}' | awk '{print $1}')) supervisor)
+ret2=$(check_process_is_running $(cat $(grep 'pid .*;' /etc/nginx/nginx.conf | awk '{print $2}' | awk -F';' '{print $1}')) nginx)
+ret3=$(check_process_is_running $(ls -l /run/tethys_asgi0.sock.lock | awk -F'-> ' '{print $2}') asgi)
+
+if [ "$ret1" != "ok" ] || [ "$ret2" != "ok" ] || [ "$ret3" != "ok" ]; then
```
swainn, line 15: Please make these `runX` variables more descriptive of what they are testing.

docker/liveness-probe.sh
```diff
@@ -0,0 +1,21 @@
...
+	echo "liveness-probe.sh: at least 1 test failed"
```
swainn, line 16: Make this message more descriptive: which tests failed?

swainn (commented):
See comments.

docker/liveness-probe.sh
gronka: I can do that; I prefer that as well

gronka:
@swainn I made the output more descriptive, and I added checks for cases that could create silent failures. Let me know if it suits you

swainn (approved):

Issue #1042, opened 2024-04-26
[BUG] Issue with MULTIPLE_APP_MODE=False after logging out


swainn:
**Describe the bug**
We encountered error after logging out single app mode:

![no_home_error](https://github.com/tethysplatform/tethys/assets/5123221/5bac3a8b-2cc9-4251-b3f6-b9edc6f19d15)

We then tried to manually go back to the root URL, which redirected to login and produced this error:

![image (6)](https://github.com/tethysplatform/tethys/assets/5123221/5e1e92b7-c8ee-4e43-9b28-8ab6ebd01aea)

It seems to be looking for a `home` view that doesn't exist.

**To Reproduce**
1. Setup a portal with single app mode
2. Login
3. Logout

**Expected behavior**
We should be able to login after logging out

**Errors/Traceback**
See above.

**Screenshots**
See above.

ckrew:
@swainn What app was installed when this happened? I can take a look

swainn:
It's our Loone app. Might have to do with the app not having a `home` controller:

![image](https://github.com/tethysplatform/tethys/assets/5123221/e601dd06-ef1c-4e0d-abe7-16151e62c342)

swainn:
I confirmed that changing the index URL and controller to “home” fixes the issue. Tethys should get those dynamical from the app class I think.

PR #1044, opened 2024-04-29
Standalone App Non-Home Index Fixes

ckrew:
This PR looks to resolve https://github.com/tethysplatform/tethys/issues/1042. The issue is that there are many redirects and reverses that point to a hardcoded "home" url. While this works fine when the single app index is home, these redirects fail when an app index is not "home". 

The PR resolves the issue by putting conditional logic in front of the hardcoded "home" redirects. If the tethys portal is in single app mode, then the "home" redirects will point towards the app index url instead. 

Along with the hardcoded "home" fix, I have also updated additional redirects that point to the app_library url which is no longer available in standalone app mode. If the tethys portal is in single app mode, these old app_library redirects will point to more appropriate endpoints instead. 

This PR also adds a warning to users on all pages when single app mode is enabled but there are no applications installed.

image

### DIFF
--- a/tests/unit_tests/test_tethys_portal/test_context_processors.py
+++ b/tests/unit_tests/test_tethys_portal/test_context_processors.py
@@ -12,12 +12,17 @@ class TestStaticDependency(unittest.TestCase):
 
     @override_settings(MULTIPLE_APP_MODE=False)
     def test_check_single_app_mode_single(self):
-        single_app_mode = context_processors.check_single_app_mode()
+        single_app_mode, configured_single_app = (
+            context_processors.check_single_app_mode()
+        )
 
         self.assertTrue(single_app_mode)
-        self.assertTrue(single_app_mode.name == "Test App")
+        self.assertTrue(configured_single_app.name == "Test App")
 
     def test_check_single_app_mode_multiple(self):
-        single_app_mode = context_processors.check_single_app_mode()
+        single_app_mode, configured_single_app = (
+            context_processors.check_single_app_mode()
+        )
 
-        self.assertTrue(single_app_mode is None)
+        self.assertFalse(single_app_mode)
+        self.assertIsNone(configured_single_app)
--- a/tests/unit_tests/test_tethys_portal/test_context_processors.py
+++ b/tests/unit_tests/test_tethys_portal/test_context_processors.py
@@ -10,6 +10,7 @@ class TestTethysPortalContext(TestCase):
     def tearDown(self):
         pass
 
+    @override_settings(MULTIPLE_APP_MODE=True)
     def test_context_processors_multiple_app_mode(self):
         mock_user = mock.MagicMock(is_authenticated=True, is_active=True)
         mock_request = mock.MagicMock(user=mock_user)
--- a/tests/unit_tests/test_tethys_portal/test_context_processors.py
+++ b/tests/unit_tests/test_tethys_portal/test_context_processors.py
@@ -12,13 +12,12 @@ class TestStaticDependency(unittest.TestCase):
 
     @override_settings(MULTIPLE_APP_MODE=False)
     def test_check_single_app_mode_single(self):
-        single_app_mode, single_app_name = context_processors.check_single_app_mode()
+        single_app_mode = context_processors.check_single_app_mode()
 
         self.assertTrue(single_app_mode)
-        self.assertTrue(single_app_name == "Test App")
+        self.assertTrue(single_app_mode.name == "Test App")
 
     def test_check_single_app_mode_multiple(self):
-        single_app_mode, single_app_name = context_processors.check_single_app_mode()
+        single_app_mode = context_processors.check_single_app_mode()
 
-        self.assertFalse(single_app_mode)
-        self.assertTrue(single_app_name is None)
+        self.assertTrue(single_app_mode is None)
--- a/tests/unit_tests/test_tethys_portal/test_context_processors.py
+++ b/tests/unit_tests/test_tethys_portal/test_context_processors.py
@@ -3,49 +3,57 @@ from django.test import override_settings
 from tethys_portal import context_processors
 
 
-class TestStaticDependency(TestCase):
+class TestTethysPortalContext(TestCase):
     def setUp(self):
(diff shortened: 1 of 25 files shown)


swainn (approved):
Approve once tests are passing

ckrew:
This PR also adds a warning to users on all pages when single app mode is enabled but there are no applications installed.

image

tethys_apps/utilities.py
```diff
@@ -620,13 +621,18 @@ def get_configured_standalone_app():
...
-        app = TethysApp.objects.get(package=standalone_app)
-    else:
-        app = TethysApp.objects.first()
-        if not app:
-            raise ObjectDoesNotExist("No Tethys Apps have been installed")
+    try:
+        if standalone_app:
+            app = TethysApp.objects.get(package=standalone_app)
+        else:
+            app = TethysApp.objects.first()
```
swainn, line 630: What happens if the database is configured, but there is no app installed? Do these raise an exception? We need to catch those exceptions too.

tethys_apps/utilities.py
```diff
@@ -620,13 +621,18 @@ def get_configured_standalone_app():
...
+    except ProgrammingError:
+        # When MULTIPLE_APP_MODE is enabled but there is no app installed, then setting up the tethys DB fails because
+        # the code is trying to fetch data from a table that doesn't exist. Catch this exception and pass so setup can
```
swainn, line 633:
```suggestion
        # ProgrammingError: the code is trying to fetch data from a table that doesn't exist. Catch this exception and pass so setup can
```

tethys_portal/context_processors.py
```diff
@@ -30,15 +31,22 @@ def tethys_portal_context(request):
...
+def check_single_app_mode(request):
     if settings.MULTIPLE_APP_MODE:
         return False, None
     else:
-        return True, get_configured_standalone_app().name
+        configured_app = get_configured_standalone_app()
+        if not configured_app:
+            messages.warning(
+                request,
+                "MULTIPLE_APP_MODE is disabled but there is no configured tethys application.",
```
swainn, line 49:
```suggestion
                "MULTIPLE_APP_MODE is disabled but there is no Tethys application installed.",
```

tethys_apps/utilities.py
```diff
@@ -620,13 +621,18 @@ def get_configured_standalone_app():
...
-    else:
-        app = TethysApp.objects.first()
-        if not app:
-            raise ObjectDoesNotExist("No Tethys Apps have been installed")
+    try:
+        if standalone_app:
+            app = TethysApp.objects.get(package=standalone_app)
+        else:
+            app = TethysApp.objects.first()
+    except ProgrammingError:
```
sdc50, line 631:
```suggestion
    except (ProgrammingError, TethysApp.DoesNotExist):
```

tethys_portal/templates/header.html
```diff
@@ -27,11 +27,21 @@
...
+                {% url 'user:profile' as header_url  %} 
+              {% endif %}
+            {% else %}
+              {% url 'app_library' as header_url %}
+            {% endif %}
             {% if ENABLE_OPEN_PORTAL or user.is_authenticated and user.is_active %}
-              <li class="nav-item">
-                <a class="h5 me-3 mb-0 nav-link{% if request.path == app_library_url %} active" aria-current="page{% endif %}" href="{{ app_library_url }}" title="{% if single_app_name %}{{ single_app_name }}{% elif site_globals.apps_library_title %}{{ site_globals.apps_library_title }}{% else %}Apps{% endif %}">{% if single_app_name %}{{ single_app_name }}{% elif site_globals.apps_library_title %}{{ site_globals.apps_library_title }}{% else %}Apps{% endif %}</a>
-              </li>
+              {% if not single_app_mode or single_app_mode and configured_single_app.name %}
```
swainn, line 40: Move this logic to the context processor and use simple context variables like `app_library_name` and `app_library_url`. Set the values in the context processor.

tethys_portal/templates/header.html
```diff
@@ -5,7 +5,7 @@
   <nav id="header-navbar" class="navbar navbar-expand-md navbar-dark shadow px-4 py-3 mx-3" style="background-color: {{ site_globals.primary_color|default:'#0a62a9' }};">
     <div class="container-fluid">
       {% block header_brand %}
-        <a id="header-brand" class="navbar-brand" href="{% url 'home' %}">
+        <a id="header-brand" class="navbar-brand" href="{% if single_app_mode %} {% if configured_single_app %} {% url configured_single_app.index_url %}  {% else %} {% url 'user:profile' %} {% endif %} {% else %} {% url 'home' %} {% endif %}">
```
swainn, line 8: Change this to use the App Name and App icon and don't show the app library link in single app mode.

swainn (commented):
Add 15pm margin-right on the profile-admin do-hickey in the app template.

tests/unit_tests/test_tethys_portal/test_context_processors.py
```diff
@@ -1,24 +1,59 @@
...
     def setUp(self):
         pass
 
     def tearDown(self):
         pass
 
-    @override_settings(MULTIPLE_APP_MODE=False)
-    def test_check_single_app_mode_single(self):
-        single_app_mode, single_app_name = context_processors.check_single_app_mode()
+    def test_context_processors_multiple_app_mode(self):
```
swainn, line 13: Avoid assuming the values of settings. Use the settings.override here to set `MULTIPLE_APP_MODE` to `True` explicitly.

swainn (commented):
One last comment. Check your tests to make sure you aren't assuming a certain setting for `MULTIPLE_APP_MODE` like what the default is. Always set settings explicitly for tests that depend on settings.

tests/unit_tests/test_tethys_portal/test_context_processors.py
ckrew: sounds good. Ill go through and update those tests

swainn (approved):

sdc50 (approved):

Issue #1048, opened 2024-05-03
BUG - NGINX and Tethys Logs not output to container logs like we expect


swainn:
The run.sh ends with a tail follow of supervisor, nginx, and tethys logs, but only the supervisor logs seem to be coming through in kubernetes logs for the container.

https://github.com/tethysplatform/tethys/blob/9ec09bd229a788a65a2b61d89b8786236f64ba61/docker/run.sh#L110

swainn:
It's probably b/c this supervisor call is blocking: 
https://github.com/tethysplatform/tethys/blob/9ec09bd229a788a65a2b61d89b8786236f64ba61/docker/run.sh#L104

I never see the `Watching Logs` output in the text.

gronka:
@swainn You're correct. Running supervisord in the foreground is the recommended way to execute it inside a container, since as a daemon the container will execute as there is no long-lived process. The tail command can be the long-lived process, however.

I was able to allow the tail command to run by extending the container with a new container with the line:

```
RUN sed -i '/usr\/bin\/supervisord/c\  /usr/bin/supervisord' run.sh
```

This removes the `no_daemon = true` part.

However, there is a hiccup - running `tail` on `/var/log/tethys/*` will fail because passing an asterisk to tail does not work for a nonexistent directory. It seems like the `/var/log/tethys` directory is created almost immediately.

```
Hotfix in Dockerfile:
RUN sed -i '/usr\/bin\/supervisord/c\  /usr/bin/supervisord\n  sleep 10s' run.sh
```

I don't know how much shorter the sleep can safely be.

```
Solution A in run.sh:
  /usr/bin/supervisord
  sleep 10s
```

It would be better to create the directories with the correct permissions, and remove `sleep 10s`. It seems like the correct permission is `root:root`, but maybe someone with more experience can validate that.

```
Solution B in run.sh:
  mkdir /var/log/tethys
  /usr/bin/supervisord
  {original tail line}
```

Finally I don't like the current method of using an asterisk, because log rotation will generate a frenzy of output. And if we specify specific files, we don't need to create the directory beforehand:

```
Solution C in run.sh (these filenames might be incorrect - my vpn is down and we are using apache):
  /usr/bin/supervisord
  tail -qF \
    /var/log/supervisor/supervisord.log \
    /var/log/nginx/access.log \
    /var/log/nginx/error.log \
    /var/log/tethys/tethys.log
```

[...]

PR #1066, opened 2024-07-11
Fix Docker log tailing

gronka:
Change the long-lived process from supervisord to tail, and prepend tail lines with the name of the file.

### Description
Fixes this issue:
https://github.com/tethysplatform/tethys/issues/1048

### Changes Made to Code:
 - run.sh

### Related
- 

### Additional Notes
-  At startup, there will be a warning printed that some log files do not exist. They can be ignored; this seemed preferable to sending `tail` errors to `/dev/null`

### Quality Checks
 - [basic testing] New code is 100% tested
 - [*] Code has been formated
 - [N/A] Code has been linted
 - [N/A] Docstrings for new methods have been added


gronka:
I made some 2 changes recommended by @araglu. I think the variable name `PREFIX` is a much better name, and using `sed` with `--unbuffered` makes new lines output immediately, which is a significantly better experience when working locallly with docker-compose.

swainn (approved):
Looks good to me. I just updated with some changes from main that should allow the docker build to go through and the docker startup test.

swainn:
Hey @gronka I just tested this locally and I'm seeing a lot of these errors and no helpful logging:

```
run.sh: line 2: $'\r': command not found
run.sh: line 3: syntax error near unexpected token `$'{\r''
run.sh: line 3: `tail_file() {
```

Looks like some carriage return characters are causing issues maybe?

swainn:
> Looks like some carriage return characters are causing issues maybe?

That was the problem I think. I opened the file in VSCode and it was show CRLF line endings. I changed it to LF, and it replaced all the line ending and is working now.

image

swainn (commented):
Please change from CRLF to LF line endings.

gronka:
@swainn Thanks for figuring it out. I program in linux, so I don't know where an `\r` could possibly come from. Maybe it had to do with rebasing after you updated my PR to `main`, and then I tried to update my own branch to `main` - you can see I ended up duplicating commit messages. I'm a bit lost on the process here, sorry.

I don't have CRLF in my files, so I'm not sure how to proceed. I just deleted my repo and cloned it, and I don't see any CLRF.

Is this PR now a hidden branch in the tethysplatform repository?

gronka:
I usually add `* text=auto` to .gitattributes, which might be nice here. 

I think I was able to check out this PR branch with `git fetch origin pull/1066/head:fix-tail`. Is that correct? I still don't see any CRLF. `dos2unix` doesn't seem to change the file

swainn:
> I don't have CRLF in my files, so I'm not sure how to proceed. I just deleted my repo and cloned it, and I still don't see any CLRF.

Hey I think I figured out what was going on. I built the Docker image for testing using Docker Desktop in Windows, so I bet the line endings go changed when I checked it out in windows. I just tested as-is in linux and it worked great. Sorry for the confusion.

PR #1072, opened 2024-07-19
Adds `tethys start` and `tethys quickstart` commands

shawncrawley:
### Description
This code, once merged, will implement both the 'tethys start' and `tethys quickstart` commands. The `tethys start` command is just shorthand for `tethys manage start`. The `tethys quickstart` command is for first-time development tethys installations and enables users to get started with Tethys in three steps with no other prompts:

```bash
conda create -n tethys -c tethysplatform -c conda-forge micro-tethys-platform
conda activate tethys
tethys quickstart
```

The `tethys quickstart` command condenses steps 3-5 as documented in ["Getting Started"](https://docs.tethysplatform.org/en/stable/installation.html) down to this single command, with the additions of automatically installing "Hello World" application running in single app mode, opening the user's default browser to http://127.0.0.1:8000, and then starting the server.

The documentation for the "Getting Started" page has been updated herein to reflect this change.

### Changes Made to Code:
 - **tethys_cli/start_commands.py** - new file that is home to the core logic of `tethys [quick]start` commands
 - **tethys_cli/scaffold_commands.py** - `prefix` argument was added to the `tethys scaffold` CLI command to allow a user to specify the full prefix path to the install location, rather than using the current working directory
 - **tethys_cli/install_commands.py** - `do_exit` argument was added to the internal function of the `tethys install` command to specify whether `exit()` should be called at the end of the installation or not; in the case of `tethys quickstart` the script execution should not be killed after installing the new hello world app as it normally would be when running the `tethys install` via the command line. The script must go on to also start the tethys server
 - **tethys_cli/gen_commands.py** - `check_existence` argument was added to the `get_destination_path` function that allows checking a file's existence to be ignored when generating a new file (*I'm actually not sure we need this though...*)
 - **tethys_cli/__init__.py** - new start and quickstart subparsers added to tethys cli command
 - **docs/installation.rst** - completely rewritten to both reflect the new `tethys quickstart` command, but also to reduce verbosity and cut to the chase in general
 - **docs/images/getting_started/hello_world_login.png** - new screenshot used in the tutorial; it depicts the login page that will automatically open in the browser upon running `tethys quickstart`
 - **docs/images/getting_started/hello_world_app.png** - new screenshot used in the tutorial; it depicts the "Hello World" app that users will see after logging into Tethys Portal for the first time.

### Related
- 

### Additional Notes
-  

### Quality Checks 
 - [x] New code is 100% tested
 - [x] Code has been formated
 - [x] Code has been linted


tethys_cli/start_commands.py
```diff
@@ -0,0 +1,134 @@
...
+    installed_apps = get_installed_tethys_items(apps=True)
+    if not installed_apps:
+        # Automatically scaffolds and installs an app
+        app_scaffold_args = Namespace(
+            name="hello_world",
+            extension=False,
+            template="default",
+            use_defaults=True,
+            overwrite=False,
+            prefix=f"{get_tethys_home_dir()}/apps",
```
swainn, line 101: I’m on the fence with scaffolding the Hello World app in the `tethys_home_dir`. Seem like a hard place to find it… but we put the portal config in there… Any reason not to do it in the `cwd`?

tethys_cli/start_commands.py
```diff
@@ -0,0 +1,134 @@
...
+    setup_django()
+    installed_apps = get_installed_tethys_items(apps=True)
+    if not installed_apps:
+        # Automatically scaffolds and installs an app
+        app_scaffold_args = Namespace(
+            name="hello_world",
+            extension=False,
+            template="default",
+            use_defaults=True,
+            overwrite=False,
```
swainn, line 100: What happens if the Hello World app is scaffolded, but not installed (e.g.: they uninstall it but the code is still in the default location)?

tethys_cli/start_commands.py
```diff
@@ -0,0 +1,134 @@
...
+        "If being done for the first time, this will also generate a portal config file, "
+        "configure the database, and prompt to scaffold an app as well.",
+    )
+    start_parser.add_argument(
+        "-p",
+        "--port",
+        type=str,
+        help="Host and/or port on which to bind the development server.",
+    )
+    start_parser.set_defaults(func=start_command)
```
swainn, line 36: based on the help, it seems this should call `quickstart_command` too. i.e. it’s just an alias for the same function.

tethys_cli/start_commands.py
```diff
@@ -0,0 +1,134 @@
...
+                    """
+                    MULTIPLE_APP_MODE: False
+                    STANDALONE_APP: hello_world
+                """,
+                )
+            ]
+        )
+        settings_command(update_settings_args)
+
+    webbrowser.open("http://127.0.0.1:8000/")
```
swainn, line 132: Why not call `webbrowser.open()` after the `start_command`?

swainn (changes requested):

tethys_cli/install_commands.py
```diff
@@ -869,9 +869,10 @@ def validate_schema(check_str, check_list):
     )
 
 
-def successful_exit(app_name, action="installed"):
+def successful_exit(app_name, action="installed", do_exit=True):
```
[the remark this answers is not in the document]
shawncrawley: If you look at old line 874 - and not the name of the function - `successful_exit` would always exit. That's why I made the do_exit default to True, so that there's not any unintended new behavior from pre-existing code. However, I now see the silliness of having a function called `successful_exit` that has the exit as optional via `do_exit` is odd. I'm exploring now doing it the way you suggest, but rather than just pull `exit(0)` out, I'm just completely getting rid of the `successful_exit` function and just calling `write_success` everywhere `successful_exit` was called. But upon looking at it more, I'm curious why we even need `exit(0)` instead of just a `return` since everywhere it is called is inside of a function. Am I missing something there?

tethys_cli/start_commands.py
shawncrawley: Because the start_command is a blocking process that starts up the tethys server (i.e. "Quit the server with CTRL-BREAK"). I'll look into if we can inject `webbrowser.open` somewhere inside of the `python <manage_path> runserver`  command.

tethys_cli/start_commands.py
shawncrawley: Yeah, not sure what my thinking was there... I think I just wanted to standardize where it would be so it can be documented. But documenting "your current working directory" is just as simple - and more straightforward in terms of transparency. I'll change that.

tethys_cli/start_commands.py
shawncrawley: I suppose this is another argument to use the `cwd` instead of that standard place... Or is it an argument to use the standard place so I could check if the code is already there to just re-install it? Not sure... I'd lean toward the `cwd` though and then if the user had manually uninstalled the original Hello World app, then I suppose they'll get a brand new Hello World app if they run `tethys quickstart` again...

docs/images/getting_started/hello_world_app.png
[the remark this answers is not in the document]
shawncrawley: No... Sounds like I should have. I'll try to see how to fix that.

tethys_cli/scaffold_commands.py
```diff
@@ -53,8 +53,14 @@ def add_scaffold_parser(subparsers):
         action="store_true",
         help="Attempt to overwrite project automatically if it already exists.",
     )
+    scaffold_parser.add_argument(
+        "-p",
+        "--prefix",
```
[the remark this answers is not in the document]
shawncrawley: Yep, I went ahead and made that change.

tethys_cli/start_commands.py
shawncrawley: Good catch. Fixed.

tethys_cli/install_commands.py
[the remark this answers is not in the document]
shawncrawley: We discussed this in a meeting and decided that I can replace `exit(0)` with `return` as long as the tests are updated to ensure they are actually exiting in the expected place. I have now done this.

tethys_cli/start_commands.py
shawncrawley: We discussed implementing a warning and exit if `tethys quickstart` is executed and a portal config already exists. So that solves this.

tethys_cli/start_commands.py
shawncrawley: It seems that how it is now is our best option. The browser will open up prematurely to the "This site can't be reached. Localhost refused to connect" page, but within a few seconds once the server finishes starting up it will auto-redirect to the app page. If that seems unacceptable, then I'm not sure what else to do.

tethys_cli/start_commands.py
shawncrawley: @swainn I'm guessing I need to remove these original images from the project history... Is that right? It looks like that would require a force push. Should I go about this a different way?

docs/installation.rst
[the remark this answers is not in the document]
shawncrawley: Done.

docs/images/getting_started/hello_world_app.png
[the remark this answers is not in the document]
shawncrawley: I'm honestly a bit unsure of what is to be done here... I just don't have experience with `git lfs`. This is the last item remaining before this can be reviewed and merged.

tethys_cli/start_commands.py
swainn: That makes sense. No action needed.

docs/images/getting_started/hello_world_app.png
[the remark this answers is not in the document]
sdc50: I think all `.png` files in the `docs/` dir will automatically get picked up by `git lfs` due to this config option:

https://github.com/tethysplatform/tethys/blob/main/.gitattributes#L2

sdc50 (approved):
Rebase first

PR #1081, opened 2024-08-17
Implements Tethys Reactpy App Scaffold

shawncrawley:
### Description
This merge request implements a Tethys ReactPy App scaffold. I will flesh out this description soon with all of the ins and outs of it. I just wanted to finally get this in as a draft pull request to get more eyes on it.

### Changes Made to Code:
 - 

### Related
- 

### Additional Notes
-  

### Quality Checks
 - [] New code is 100% tested
 - [] Code has been formated
 - [] Code has been linted
 - [] Docstrings for new methods have been added


environment.yml
```diff
@@ -21,8 +21,7 @@ dependencies:
 
   # core dependencies
   - django>=3.2,<6
-  - channels
-  - daphne
+  - channels["daphne"]
```
sdc50, line 24: Is this valid syntax for a conda package? I though that was a `pip` thing.

tethys_apps/base/controller.py
```diff
@@ -36,6 +37,9 @@
 from typing import Union, Any
 from collections.abc import Callable
 
+global_page_controller = optional_import(
```
sdc50, line 40: Why is this an `optional_import`?  Since it's part of the source code we should be able to import it directly right?

environment.yml
shawncrawley: Hm... It must be, since I installed tethys fresh from this branch on a new machine a couple of days ago and it worked without issue.

tethys_apps/base/controller.py
shawncrawley: I debated a bit on how I should implement this. It's optional because the tethys_apps.base.page_handler module requires that `reactpy` (the true optional import) be imported at the top level since one of the importable functions itself has a `reactpy` decorator. So importing anything from `tethys_apps.base.page_handler` would fail if `reactpy` was not installed. But I just dug into it again, and realized that I actually updated the `reactpy` import in `tethys_apps.base.page_handler` itself to have optional behavior. So I think you're right that we ultimately don't need the optional behavior here anymore. I'll get that cleaned up.

tethys_apps/base/url_map.py
```diff
@@ -57,14 +61,16 @@ def __init__(
...
+        self.index = index
 
     def __repr__(self):
         """
         String representation
         """
         return (
             f"<UrlMap: name={self.name}, url={self.url}, controller={self.controller}, protocol={self.protocol}, "
-            f"handler={self.handler}, handler_type={self.handler_type}>"
+            f"handler={self.handler}, handler_type={self.handler_type}, title={self.title}, index={self.index}>"
```
sdc50, line 73: Is this only being used by the ReactPy stuff, or is it/can it be leveraged more broadly?

tethys_apps/templatetags/site_settings.py
```diff
@@ -14,23 +14,27 @@
 @register.filter
 @stringfilter
 def load_custom_css(var):
-    if var.startswith("/"):
-        var = var.lstrip("/")
+    if var.startswith(os.sep):
```
sdc50, line 17: Does this work on Windows? If a file path is absolute on Windows it will start with something like `C:\` right?

sdc50, line 17: I'm not sure that this is the right move here. I'm trying to decide if the user should specify these as a URL path or a system path.

tethys_apps/templatetags/site_settings.py
```diff
@@ -14,23 +14,27 @@
...
 
     is_file = os.path.isfile(
         os.path.join(settings.STATIC_ROOT, var)
     ) or static_finder.find(var)
 
     if is_file:
-        return '<link href="' + os.path.join("/static", var) + '" rel="stylesheet" />'
+        return (
+            '<link href="'
+            + os.path.join(os.sep, "static", var)
```
sdc50, line 27: In this case these are definitely URL paths so this approach won't work. To ensure we are supporting windows we'll need to ensure these are output as Posix Paths.

tethys_cli/install_commands.py
```diff
@@ -855,9 +856,11 @@ def install_command(args):
         if validate_schema("post", install_options):
             write_msg("Running post installation tasks...")
             for post in install_options["post"]:
-                path_to_post = file_path.resolve().parent / post
+                command = file_path.resolve().parent / post
```
sdc50, line 859: I would prefer this to be called something like `script` rather than `command`. Since it is a path to a file I think `command` is somewhat missleading.

tethys_cli/scaffold_templates/app_templates/reactpy/install.yml_tmpl
```diff
@@ -0,0 +1,19 @@
...
+
+requirements:
+  # Putting in a skip true param will skip the entire section. Ignoring the option will assume it be set to False
+  skip: false
+  conda:
+    channels:
+    packages:
+
+  pip:
+    - reactpy-django
```
sdc50, line 16: I'd love to see this get put on `conda-forge`

tethys_cli/scaffold_templates/app_templates/reactpy/setup.py_tmpl
```diff
@@ -0,0 +1,31 @@
+from setuptools import setup, find_namespace_packages
```
sdc50, line 1: I think we're going to need to move to `pyproject.toml` sooner that later. Should we just start this scaffold out that way so we don't have to migrate it later?

tethys_portal/settings.py
```diff
@@ -220,7 +220,7 @@
 )
 
 default_installed_apps = [
-    "channels",
```
sdc50, line 223: channels is not needed?

sdc50 (commented):
@shawncrawley This is amazing and I'm really excited to get it officially in Tethys. I have several questions (see inline). 

In general I would prefer that `pathlib.Path` be used rather than `os.path`, but I won't hold this up for that if you don't have the bandwidth to change that now.

tethys_apps/base/url_map.py
shawncrawley: This is only being used for ReactPy stuff. I did not put any though into whether it could be used more broadly. This was implemented specifically for use in the automatic generation of the navigation bar links - one per function decorated by `@page`.

tethys_cli/install_commands.py
shawncrawley: Yeah, I'm honestly not sure why I even switch to `command` from `path_to_post`. I think I'll just revert to `path_to_post`.

tethys_cli/scaffold_templates/app_templates/reactpy/post_install.py
```diff
@@ -0,0 +1,8 @@
+from tethys_cli.settings_commands import read_settings, write_settings
```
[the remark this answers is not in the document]
shawncrawley: Oh wow, that's a great idea. I would much prefer that, I just didn't even think of it.

tethys_cli/scaffold_templates/app_templates/reactpy/setup.py_tmpl
shawncrawley: That makes sense to me. I don't know much about pyproject.toml, but my guess is it'd be a quick learn, yeah? I'll make it happen.

tethys_portal/settings.py
shawncrawley: According to the latest documentation, it appears it is not necessary unless using Channel's runworker command (see [here](https://channels.readthedocs.io/en/stable/installation.html)). However, this likely isn't the case on the previous versions of Django. Since including it seems to be optional now, we should probably include it anyway to get backwards compatibility.

tethys_apps/templatetags/site_settings.py
sdc50: This is what we should be using: `urllib.parse.ulrjoin` https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

shawncrawley:
@sdc50 I implemented your feedback. In full disclosure, I got a little carried away when swapping out os.path with pathlib.Path. I ended up removing os.path from the entire project, except for about two spots where pathlib.Path was actually more clunky or problematic. So this PR is definitely super gold-plated by scope creep at this point. That means that there are a lot of new modified files since your last review... I apologize in advance. I was already in too deep before I determined that I shouldn't have added all of that to this specific PR.

tests/unit_tests/test_tethys_apps/test_base/test_workspace.py
```diff
@@ -42,72 +42,72 @@ class TethysAppChild(tethys_app_base.TethysAppBase):
...
+        if self.test_root.is_dir():
             shutil.rmtree(self.test_root)
-        if os.path.isdir(self.test_root2):
+        if self.test_root2.is_dir():
             shutil.rmtree(self.test_root2)
 
     def test_TethysWorkspace(self):
         # Test Create new workspace folder test_workspace
-        result = base_workspace.TethysWorkspace(path=self.test_root)
+        result = base_workspace.TethysWorkspace(path=str(self.test_root))
```
sdc50, line 60: Doesn't `TethysWorkspace` accept a `Path` object?

tests/unit_tests/test_tethys_apps/test_template_loaders.py
```diff
@@ -58,13 +59,16 @@ def test_get_contents_template_does_not_exist(self, _, mock_file):
...
-        mock_gdt.return_value = ["/foo/template1"]
+        mock_gdt.return_value = [str(Path.home() / "foo" / "template1")]
         expected_template_name = "foo"
 
         for origin in tethys_template_loader.get_template_sources(
             expected_template_name
         ):
-            self.assertEqual("/foo/template1/foo", origin.name)
+            self.assertEqual(
+                str(Path(Path.home() / "foo" / "template1" / "foo")),
```
sdc50, line 69: no need to use `Path` twice here.

tethys_cli/cli_helpers.py
```diff
@@ -41,19 +41,19 @@ def get_manage_path(args):
     Validate user defined manage path, use default, or throw error
     """
     # Determine path to manage.py file
-    manage_path = os.path.join(get_tethys_src_dir(), "tethys_portal", "manage.py")
+    manage_path = f"{get_tethys_src_dir()}/tethys_portal/manage.py"
```
sdc50, line 44: Why not use `Path` here?

tests/unit_tests/test_tethys_apps/test_base/test_workspace.py
sdc50: So `TethysWorkspace` will accept a `Path`. With your changes it just converts it to a `Path` first think anyway. However, it's not worth changing since this works fine and the workspaces API is going to be dropped in 5.0.

tests/unit_tests/test_tethys_apps/test_template_loaders.py
sdc50, line 69:
```suggestion
                str(Path.home() / "foo" / "template1" / "foo"),
```

tethys_cli/cli_helpers.py
sdc50, line 44:
```suggestion
    manage_path = Path(get_tethys_src_dir()) / "tethys_portal" / "manage.py"
```

sdc50 (changes requested):
@shawncrawley just a couple of minor changes (I've added suggestions that you can just commit if you agree with them).

I'm very excited about the reactpy addition. The `Path` changes are all internal, but everything is converted back to a string in return values. I think this is the right call for now (to ensure we aren't breaking anything). However, in Tethys 5.0 I think we should switch and return `Path` objects wherever possible.

tethys_cli/cli_helpers.py
shawncrawley: Good catch! I'll fix that.

shawncrawley:
@swainn @sdc50 I've got everything done and passing tests except oddly the "Tethys-CI / Conda Build (ubuntu-latest)" test run. Any thoughts there on what could be going on? I have reactpy-django now installed by default, which did throw a wrench in some of the testing. Maybe it's related here? But oddly only this one build...

shawncrawley:
@gagelarsen Do you have any idea on the failing "Conda Build" GitHub workflow?

tethys_cli/scaffold_templates/app_templates/reactpy/pyproject.toml_tmpl
swainn, line 1: 💖 Update the other scaffolds to use this pyrpoject.toml too, please.

swainn: I think you need to add something like this to include non-python files in the package build:

```
[tool.setuptools]
include-package-data = true

[tool.setuptools.packages.find]
include = ["tethysapp*"]

[tool.setuptools.package-data]
"*" = [
    "*.js",
    "*.png",
    "*.gif",
    "*.jpg",
    "*.html",
    "*.css",
    "*.gltf",
    "*.json",
    "*.svg",
]
```

swainn (changes requested):

environment.yml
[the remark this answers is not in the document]
shawncrawley: Should I still list "reactpy" separately and distinctly, even though it's baked in as an underlying requirement of "reactpy-django"?

environment.yml
```diff
@@ -105,3 +104,7 @@ dependencies:
   - factory_boy
   - flake8
   - flake8-bugbear
+
+  # reactpy dependencies
+  - pip:
+    - reactpy-django
```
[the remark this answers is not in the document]
shawncrawley: I've been looking into that. Unfortunately, it looks like there may be 4 total packages we'd need to add, since all of the dependencies also need to be in conda-forge. Those 4 packages are: reactpy-django, reactpy, reactpy-router, and nodejs-bin...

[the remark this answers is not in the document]
swainn: That's pretty typical for adding conda packages. I think you might be able to add them all in one go if they are related...

tethys_sdk/components/utils.py
```diff
@@ -0,0 +1,2 @@
+from reactpy import component, event as event_decorator  # noqa: F401
```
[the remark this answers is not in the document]
shawncrawley: Yeah, we could just leave it as event. In the cases that I used it, I would end up wanting `event` to be an argument of some of the functions used surrounding it. I just didn't want any namespace confusion. But it would make more sense to rename my function args rather than the decorator.

tests/unit_tests/test_tethys_apps/test_base/test_app_base.py
```diff
@@ -1543,3 +1544,36 @@ def test_remove_from_db_2(self, mock_ta, mock_log):
...
+        self.assertListEqual(links, app.nav_links)
+
+    def test_navigation_links_auto_excluded_page(self):
+        app = tethys_app_base.TethysAppBase()
+        app.nav_links = "auto"
+        app.index = "home"
+        app.root_url = "test-app"
+
+        app._registered_url_maps = [
+            Namespace(name="exclude_page", title="Exclude Page", index=-1),
```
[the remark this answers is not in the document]
shawncrawley: I just wanted a mock-like instance of UrlMap... I should have just used a mock.MagicMock object, lol. I'll swap it for that.

environment.yml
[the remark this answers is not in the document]
shawncrawley: I went ahead and added `reactpy` to the dependencies

environment.yml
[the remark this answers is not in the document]
shawncrawley: I went ahead and added `reactpy` to the dependencies.

tests/unit_tests/test_tethys_apps/test_base/test_app_base.py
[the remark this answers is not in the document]
shawncrawley: Turns out I must've chosen `Namespace` for a reason. Using `mock.MagicMock` causes the tests to fail because mocking the property `name` cannot be done on initialization (see [mock docs](https://docs.python.org/3/library/unittest.mock.html#mock-names-and-the-name-attribute)). While I could have broken it out into two steps, it seemed like using `Namespace` shouldn't be a big deal. Let me know if you disagree.

tethys_cli/scaffold_templates/app_templates/reactpy/pyproject.toml_tmpl
shawncrawley: Done

shawncrawley:
It looks like the only thing that remains is getting `reactpy` and `reactpy-django` added to conda-forge. I found that we'd also have to add `reactpy-router`, which unfortunately doesn't have an sdist file (required, I believe, by conda-forge). But I reached out to the reactpy maintainers and they quickly responded that they'll get the sdist file added by tomrorow. Once that's done, I should be able to submit the PR to the [conda staged-recipes](https://github.com/conda-forge/staged-recipes) repo and get those added. It's the final stretch!

tests/unit_tests/test_tethys_apps/test_base/test_app_base.py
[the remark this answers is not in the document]
swainn: Try using the `UrlMaps` for the test instead of the `Namespace`.

swainn (approved):
Barring conda package and all comments addressed.

PR #1084, opened 2024-08-23
Fix in scaffold for extension

jakeymac:
### Description
This merge fixes an import in the default setup.py file that is generated when scaffolding a tethys extension


swainn:
@jakeymac This PR has two commits. The first changes the import. The second changes it back... As a result there are no changes. plz fix

swainn (approved):

Issue #1069, opened 2024-07-18
[FEATURE] Can we remove SUPERUSER from the PostgreSQL create user command?


gronka:
I would like to use Azure PostgreSQL, but if Tethys requires the SUPERUSER role to work, then it's not possible.

**Is your feature request related to a problem? Please describe.**
Azure has a PostgreSQL service which does not allow the creation of a SUPERUSER. Azure keeps that role for themselves. I imagine other PostgreSQL PaaS would make the same decision.

**Describe the solution you'd like**
Remove the `SUPERUSER` assignment from line 281 of `tethys_cli/db_commands.py`. Tethys seems to run fine if the superuser is not actually a superuser - but I could easily be missing something.

**Describe alternatives you've considered**
The alternative offered by Tethys is to set `SKIP_DB_SETUP=true`, and Tethys seems to run fine after running the following commands with psql:

```
CREATE EXTENSION IF NOT EXISTS postgis;
CREATE USER tethys_super WITH CREATEDB CREATEROLE PASSWORD 'pass';
CREATE USER tethys_default WITH NOCREATEDB NOCREATEROLE NOSUPERUSER PASSWORD 'pass';
CREATE DATABASE tethys_platform WITH OWNER tethys_default TEMPLATE template0 ENCODING 'UTF8';
```

mattw-nws:
Similar issue on AWS RDS, 👍 

Things can/will fail if the `tethys_super` user can't for example create tables and databases (this can be worked around by having all that precreated, but that's likely to be fragile and will I think prevent things such as migrations from working). But no, it doesn't have to be `SUPERUSER` strictly speaking.

I worked around it by pre-creating the `tethys_super` user with the `rds_superuser` instead, similar to below:

```
# This has to happen for RDS to work, and if it fails, we must not be using RDS...
#TODO: Any reason to be concerned about injection here? Just strip all non-word characters?
psql -h ${TETHYS_DB_HOST} -U postgres -c "CREATE ROLE ${TETHYS_DB_SUPERUSER} NOSUPERUSER IN ROLE rds_superuser INHERIT CREATEDB CREATEROLE NOREPLICATION VALID UNTIL 'infinity' LOGIN PASSWORD '${TETHYS_DB_SUPERUSER_PASS}'" \
|| true
```

(The `|| true` is a bit of a hack but is minimally sufficient for this to be runable in a script whether I'm on RDS or a local dev env.)

[...]

PR #1087, opened 2024-08-28
Nonexit Tethys Create DB CLI Command

ckrew:
### Description
This merge request addresses an issue with the tethys db create CLI command. The issue occurs when multiple portals are trying to share a single database with the same tethys_super table. If the tethys_super table already exists, then the tethys db create CLI command will try to create the tethys_super table, fail to do so, and then exit before the table for the portal is created. 

This merge request sets the exit_on_error to False for the create_db_user call inside of the create_tethys_db function. This will ensure that both create_db_user calls are executed, whether they are successful or not. To keep the same behavior currently where the create_tethys_db function can exit on an error, I have added a exit_creation_on_error argument to the function. If set to True, then the function will exit if either one of the create_db_user calls fails.

I also updated the logging in the create_tethys_db function to give more details on what user or database is being created

### Steps to reproduce
- Setup the tethys_super DB table (without having the tethys default/portal table setup)
- Run tethys db create
- You will receive an error that the tethys_super table already exists and the tethys default table will not be created

### Quality Checks
 - [X] New code is 100% tested
 - [X] Code has been formated
 - [X] Code has been linted
 - [X] Docstrings for new methods have been added

### DIFF
--- a/tests/unit_tests/test_tethys_cli/test_db_commands.py
+++ b/tests/unit_tests/test_tethys_cli/test_db_commands.py
@@ -197,18 +197,63 @@ class TestCommandTests(unittest.TestCase):
             **kwargs,
         )
 
+    @mock.patch("tethys_cli.db_commands.write_error")
     @mock.patch("tethys_cli.db_commands.create_db_user")
-    def test_db_command_create(self, mock_create_db_user):
+    def test_db_command_create_error_code_without_exit(
+        self, mock_create_db_user, mock_write_error
+    ):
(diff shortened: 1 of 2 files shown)


ckrew:
Removing the "SUPERUSER" role as part of this PR to cover #1069

ckrew:
> Removing the "SUPERUSER" role as part of this PR to cover #1069

This actually caused the tests to fail so I added it back in so we could get this merged.

swainn (approved):

PR #1102, opened 2024-10-07
Delayed job status

sdc50:
### Description
Add the ability to add a query argument `delay` to the `update_job_status` callback, e.g.:

```
http://127.0.0.1:8000/update-job-status/14/?delay=120
```

The server will wait for `delay` seconds before calling `update_status` on the job.

#### Note: 
This "delay" capability was added as a general capability, and the ability schedule periodic execution of a function was included (though not used by the `update_job_status` callback). I have not added this to the Tethys SDK because I would like to get more experience with it before deciding if it should be made "officially" available to apps. I don't consider it to be a robust task scheduling capability since all tasks are stored in memory and would be lost if the server restarts.

### Changes Made to Code:
 - modified the `update_job_status` callback to be async and to support a delay
 - moved the job status views from `tethys_apps` to `tethys_compute`
 - migrated a few helper functions from `tethys_gizmos` to `tethys_compute`

### Quality Checks
 - [x] New code is 100% tested
 - [x] Code has been formated
 - [x] Code has been linted

### DIFF
--- a/tethys_compute/tasks.py
+++ b/tethys_compute/tasks.py
@@ -15,6 +15,22 @@ logger = logging.getLogger(f"tethys.{__name__}")
 
 
 def create_task(func, /, *args, delay=0, periodic=False, count=None, **kwargs):
+    """
+    Schedules a task to be executed after some delay. This is run asynchronously and must be called from a context
+    where there is an active event loop (e.g. from a controller).
+
+    Can be set to run periodically (i.e. it will be rescheduled after it is run) either indefinitely or for a
+    specified number of times.
+    Args:
+        func (callable): the function to schedule
+        *args: args to pass to the function when it is called
+        delay (int): number of seconds to wait before executing `func` or between calls if `periodic=True`
+        periodic (bool): if `True` the function will be rescheduled after each execution until `count=0` or
+            indefinitely if `count=None`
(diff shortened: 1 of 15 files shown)


sdc50:
I'm not sure what happened to `readthedocs`, but our build there seems to be broken. I'm out of time to look into it for now.

swainn:
> I'm not sure what happened to `readthedocs`, but our build there seems to be broken. I'm out of time to look into it for now.

Best I can tell, what happened is Python 3.13 was released and the `cgi` module that Django 3.2 depended on was removed (previously deprecated). Our docs build was pinned to django=3.2.*. I just updated it to 4.2.* and the build worked.

tethys_compute/tasks.py
```diff
@@ -0,0 +1,73 @@
...
+            await result
+    except Exception as e:
+        logger.info(
+            f'The following error occurred while running the task "{func}": {e}'
+        )
+    if periodic and (count is None or count > 0):
+        if isinstance(count, int):
+            count -= 1
+        asyncio.create_task(
+            _run_after_delay(
```
swainn, line 70: The other limitation of this approach is that eventually you will hit the max recursion depth on an indefinite task. In the future, we should look at integrating with something like cron or celery.

swainn (approved):

PR #1107, opened 2024-10-28
Updated ENABLE_RESTRICTED_APP_ACCESS for potential list value

ckrew:
### Description
Currently the ENABLE_RESTRICTED_APP_ACCESS is set to be a boolean. If False, all apps are accessible to everyone. If True, then no apps are accessible unless given explicit permissions. This causes an issue if I only want 1 or a few apps restricted and leave the rest accessible to everyone.

This PR aims to fix that by maintaining previous capabilities of a boolean value but also adding the option to provide a list of restricted applications. If an application is provided in the list, then that application will only be accessible by admins and those given explicit permissions. 

### Changes Made to Code:
 - Added logic to the existing utilities function that checks for apps in a restricted list if the setting is a list

### Additional Notes
-  No expected results had to be changed in tests which means it should not cause any breaking issues in current portals

### Quality Checks
 - [X] New code is 100% tested
 - [X] Code has been formated
 - [X] Code has been linted


swainn (approved):
LGTM, but I want @sdc50 to review and comment since the impetus for this feature originally came from his needs.

sdc50:
I don't have a direct need for this capability currently, but I can see how it could be useful. It seems to work well for me with both native and proxy apps.

PR #1109, opened 2024-10-29
Flexible Conda Dependencies

sdc50:
- Fixes the `install_tethys.sh` script so the specified version of `python` is installed
- Avoids modifying the `environment.yml` file so the conda packages will have flexible dependencies
- Modifies some syntax in the codebase to be compatible with `python=3.10`


sdc50:
image

I'm not sure how the `python=3.10` tests were passing previously.

sdc50:
> I'm not sure how the `python=3.10` tests were passing previously as we use this syntax:

It appears that the `python` version wasn't being set correctly in the test matrix previously: https://github.com/tethysplatform/tethys/actions/runs/11576539437/job/32285872638?pr=1107#step:4:264

The `sed` command to substitute the `python` version had a typo, so it wasn't actually pinning the python version.

The good news is that this PR fixes that! The bad news is that our code base is not currently compatible with `python=3.10` like with thought.

sdc50:
It looks like all checks are passing with the current approach (create the environment with the original `environment.yml` file and then afterward install the correct version of `python` and `django`). However, when I did this locally (downgrading to `python=3.10`) it tried to install `sqlalchemy=2` and also was going to downgrade `tethys_dataset_services`. I checked the test environment on GitHub Actions and it looks like it didn't do this, but I still think it prudent to modify the approach. 

For both efficiency and stability I'm going to modify the contents of the `environment.yml` file, but rather than overwrite the original file, it write it out to a new file. The new file will be used to create the test environment, but the original `environment.yml` file will remain unchanged so that it can be used to create the correct (flexible) dependencies for the conda package.

scripts/install_tethys.sh
```diff
@@ -363,25 +363,13 @@ then
...
-        sudo sed -i.bak "s/django>=.*/python==${PYTHON_VERSION}/" "${TETHYS_SRC}/environment.yml"
-        sudo sed -i.bak "s/django>=.*/python==${PYTHON_VERSION}/" "${TETHYS_SRC}/micro_environment.yml"
-    fi
-
     if [ -n "${CREATE_ENV}" ]
     then
         # create conda env and install Tethys
         echo "Setting up the ${CONDA_ENV_NAME} environment..."
-        conda env create -n ${CONDA_ENV_NAME} -f "${TETHYS_SRC}/environment.yml"
+        sudo sed "s/python>=.*/python=${PYTHON_VERSION}/" "${TETHYS_SRC}/environment.yml" > "${TETHYS_SRC}/temp_environment.yml"
```
gagelarsen, line 370: Maybe change this to `generated_environment.yml` or something like that? Unless we are just getting rid of it? Temp just makes me think we need to get rid of it when we are done. But might be nice to keep it around so if we have any question as to what was used to generate the environment.

sdc50: This seems like a reasonable change. While the `temp_environment.yml` file is only needed during the environment setup, and, as far as I am aware, the only use case for the `install_tethys.sh` file is in the CI, it is conceivable that it would be used in another situation in which case `generated_environment.yml` file would more self-documenting.

gagelarsen (approved):
Looks good to me.

PR #1120, opened 2024-11-28
4.3 What's New

sdc50:
This updates the `whats_new.rst` doc for the 4.3 release. It also has:

- a quick fix for the docs build to work with Python 3.13
- adds docs for the new `CONTEXT_PROCESSORS` setting
- updates the version in  `CITATION.cff` file 
  - The [doi link](https://doi.org/10.5281/zenodo.5228298) in this file points to a 3.3 version of Tethys. I'm not sure if/how we can update that.


swainn (approved):

sdc50:
I think the Python `conda` package is not yet compatible with Python 3.13, so the full environment won't resolve with that version of Python yet.

swainn:
It looks like Path.walk was added in Python 3.12. We’ll need to use os.walk I think to support back to 3.10.

swainn:
Merged!

Issue #1127, opened 2024-12-05
[BUG] Long custom styles cause OSError 36 File name too long


araglu:
**Describe the bug**
Very long CSS text in Custom Styles causes a "File name too long" error. I had 498 characters in "Base Portal CSS", and that caused the error to show in the admin pages which made it difficult to remove the CSS text.

I think this problem was caused by the switch from os.path.isfile() to Pathlib .is_file() over in https://github.com/tethysplatform/tethys/blob/main/tethys_apps/templatetags/site_settings.py#L20. I never would have thought Pathlib would behave differently there.

I could not duplicate this problem in Windows. This only breaks in Linux for me.

**To Reproduce**
1. Go to Site Admin -> Tethys Portal -> Site Settings -> Custom Styles
2. Enter in 257-500 characters of CSS in the "Portal Base CSS" field.
3. Click SAVE.
4. See an immediate "500 Internal Server Error". The server log shows "OSError: [Errno 36] File name too long:"

**Errors/Traceback**
This Traceback comes from a new Tethys environment in WSL using `tethys quickstart`.
<details><summary>Details</summary>
<p>

```
System check identified no issues (0 silenced).
December 05, 2024 - 19:11:16
Django version 5.1.4, using settings 'tethys_portal.settings'
Starting ASGI/Daphne version 4.1.2 development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
ERROR Internal Server Error: /admin/tethys_config/settingscategory/
Traceback (most recent call last):
  File "/home/lugarmd/mambaforge/envs/tethys43-plain/lib/python3.12/site-packages/asgiref/sync.py", line 518, in thread_handler
    raise exc_info[1]
  File "/home/lugarmd/mambaforge/envs/tethys43-plain/lib/python3.12/site-packages/django/core/handlers/exception.py", line 42, in inner
    response = await get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lugarmd/mambaforge/envs/tethys43-plain/lib/python3.12/site-packages/asgiref/sync.py", line 518, in thread_handler
    raise exc_info[1]
  File "/home/lugarmd/mambaforge/envs/tethys43-plain/lib/python3.12/site-packages/django/core/handlers/base.py", line 284, in _get_response_async
    response = await sync_to_async(
               ^^^^^^^^^^^^^^^^^^^^
[...]

PR #1128, opened 2024-12-05
Avoid "File name too long" error

araglu:
### Description
This merge request addresses https://github.com/tethysplatform/tethys/issues/1127 where CSS text greater than 256 characters causes an OSError exception on many pages, including admin pages.

### Changes Made to Code:
 - On any file error, treat "var" as CSS text to embed
 - Add docstring
 - Reorder imports
 - Remove unnecessary else: after return
 - Add unit test

### Related
- https://github.com/tethysplatform/tethys/issues/1127

### Additional Notes
- While making the unit test, I found that added a /* CSS Comment */ bypassed the issue. So the test variable is a little shorter than I hoped for. Also, that test CSS is not going to win any design awards.

### Quality Checks
 - [x] New code is 100% tested
 - [x] Code has been formatted (black)
 - [x] Code has been linted (pylint, flake8)

### DIFF
--- a/tethys_apps/templatetags/site_settings.py
+++ b/tethys_apps/templatetags/site_settings.py
@@ -1,3 +1,4 @@
+import logging
 from pathlib import Path
 
 from django import template
@@ -10,6 +11,8 @@ static_finder = TethysStaticFinder()
 
 register = template.Library()
 
+log = logging.getLogger(f"tethys.{__name__}")
+
 
 @register.filter
 @stringfilter
@@ -19,9 +22,12 @@ def load_custom_css(var):
     Args:
         var: a filename of CSS to load or CSS text to embed into the page
 
-    Returns: a string of HTML that either embeds CSS text or points to a file
+    Returns:
+        a string of HTML that either embeds CSS text or points to a file
 
     """
+    if not var.strip():
+        return ""
     if var.startswith("/"):
         var = var.lstrip("/")
 
@@ -32,7 +38,18 @@ def load_custom_css(var):
         for path in settings.STATICFILES_DIRS:
             if (Path(path) / var).is_file():
                 return f'<link href="/static/{var}" rel="stylesheet" />'
-    except OSError:
+    except OSError as e:
+        oserror_exception = ": " + str(e)
         pass
+    else:
+        oserror_exception = ""
+
+    common_css_chars = "{};,"
+    if not any(c in var for c in common_css_chars):
+        # This appears to be a filename and not a CSS string
+        log.warning(
+            "Could not load file '%s' for custom styles%s", var, oserror_exception
+        )
+        return ""
 
     return "<style>" + var + "</style>"
--- a/tethys_apps/templatetags/site_settings.py
+++ b/tethys_apps/templatetags/site_settings.py
@@ -39,7 +39,7 @@ def load_custom_css(var):
         for path in settings.STATICFILES_DIRS:
             if (Path(path) / var).is_file():
                 return f'<link href="/static/{var}" rel="stylesheet" />'
-    # If the string is too long for a file path, which could happen if it is CSS, 
+    # If the string is too long for a file path, which could happen if it is CSS,
     # an OSError will be raised during the file path checks. This could also happen
     # if a lengthy file path is given or is otherwise invalid.
     except OSError as e:
--- a/tethys_apps/templatetags/site_settings.py
+++ b/tethys_apps/templatetags/site_settings.py
@@ -1,8 +1,8 @@
+from pathlib import Path
+
 from django import template
 from django.template.defaultfilters import stringfilter
 from django.conf import settings
-
-from pathlib import Path
 
 from ..static_finders import TethysStaticFinder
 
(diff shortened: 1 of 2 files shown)


tethys_apps/templatetags/site_settings.py
```diff
@@ -14,15 +14,25 @@
...
+    try:
+        if (Path(settings.STATIC_ROOT) / var).is_file() or static_finder.find(var):
+            return f'<link href="/static/{var}" rel="stylesheet" />'
 
-    else:
         for path in settings.STATICFILES_DIRS:
             if (Path(path) / var).is_file():
                 return f'<link href="/static/{var}" rel="stylesheet" />'
+    except OSError:
+        pass
```
swainn, line 36: Is the assumption here that if the sting is too long, an OSError will be thrown and therefore it's not a file path, so it must be a css string? What about paths that really are too long? It seems like we should make our checking logic more robust and validate that the string is css if the OSError is raised and if not, log a warning.

swainn (commented):
Thanks for the patch @araglu. See my comment.

tethys_apps/templatetags/site_settings.py
araglu: That's what I was going for. I also thought about specifically targeting Errno 36, but any OSError should steer it towards a CSS string.

I will add something here to check if `var` has any curly braces or semicolons or commas, and then log an error if not. Something like this plus setting up logger and testing it:
```suggestion
        common_css_chars = "{};,"
        if not any(c in var for c in common_css_chars):
            # This appears to be a filename and not a CSS string
            logger.warning("Could not load file '%s' for custom styles. %s", var, str(e))
            return ""
```

swainn: That seems like a reasonable approach for detecting the CSS. Whatever feedback we can provide to the user/administrator would be great.

araglu: - Added filename detection code and log warning to check all input that isn't a file. This should help show problems with a missing file or misspelled filename, not just the rare thing that could trigger an OSError.
- load_custom_css() gets called a lot, so I added a quick return if `var` is empty. That also prevents false positives with the log warning.
- Updated test_get_css_is_code() to test with CSS code instead of "test.css", since load_custom_css() now returns an empty string if it detects a filename that can't be found.

tethys_apps/templatetags/site_settings.py
```diff
@@ -1,28 +1,55 @@
...
+        if (Path(settings.STATIC_ROOT) / var).is_file() or static_finder.find(var):
+            return f'<link href="/static/{var}" rel="stylesheet" />'
 
-    else:
         for path in settings.STATICFILES_DIRS:
             if (Path(path) / var).is_file():
                 return f'<link href="/static/{var}" rel="stylesheet" />'
+    except OSError as e:
+        oserror_exception = ": " + str(e)
+        pass
```
swainn, line 43:
```suggestion
```

tethys_apps/templatetags/site_settings.py
```diff
@@ -1,28 +1,55 @@
...
+    if not var.strip():
+        return ""
     if var.startswith("/"):
         var = var.lstrip("/")
 
-    if (Path(settings.STATIC_ROOT) / var).is_file() or static_finder.find(var):
-        return f'<link href="/static/{var}" rel="stylesheet" />'
+    try:
+        if (Path(settings.STATIC_ROOT) / var).is_file() or static_finder.find(var):
+            return f'<link href="/static/{var}" rel="stylesheet" />'
```
swainn, line 36:
```suggestion
    try:
        # Check if var is a path to a file, if so return a link tag to the file
        if (Path(settings.STATIC_ROOT) / var).is_file() or static_finder.find(var):
            return f'<link href="/static/{var}" rel="stylesheet" />'
```

tethys_apps/templatetags/site_settings.py
```diff
@@ -1,28 +1,55 @@
...
 
-    else:
         for path in settings.STATICFILES_DIRS:
             if (Path(path) / var).is_file():
                 return f'<link href="/static/{var}" rel="stylesheet" />'
+    except OSError as e:
```
swainn, line 41:
```suggestion
    # If the string is too long for a file path, which could happen if it is CSS, 
    # an OSError will be raised during the file path checks. This could also happen
    # if a lengthy file path is given or is otherwise invalid.
    except OSError as e:
```

tethys_apps/templatetags/site_settings.py
```diff
@@ -1,28 +1,55 @@
...
+        oserror_exception = ": " + str(e)
+        pass
+    else:
+        oserror_exception = ""
+
+    common_css_chars = "{};,"
```
swainn, line 47:
```suggestion
    # Verify the string is CSS and log warning if it is not
    common_css_chars = "{};,"
```

swainn (changes requested):
A few suggestions for comments to aid future devs. Feel free to refine/correct.

araglu:
I like all those changes. Thanks!

swainn (approved):

swainn:
All merged. Thank you for the contribution @araglu!

PR #1158, opened 2025-02-18
Quotas bug fix

jakeymac:
This pull request addresses a few bugs involving workspace quota enforcement. 

- fixes workspace quotas not being enforced
- fixes workspace management not being possible due to workspace quotas (viewing/clearing current workspace usage)

### DIFF
--- a/tethys_apps/base/paths.py
+++ b/tethys_apps/base/paths.py
@@ -364,6 +364,7 @@ def _get_user_workspace(app_or_request, user_or_request, bypass_quota=False):
 def get_user_workspace(
     app_or_request,
     user_or_request,
+    bypass_quota=False,
 ) -> TethysPath:
     """
     Get the dedicated user workspace for the given app. If an HttpRequest is given, the workspace of the logged-in user will be returned (i.e. request.user).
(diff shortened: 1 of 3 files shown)


tethys_portal/settings.py
```diff
@@ -623,7 +623,7 @@ def get__all__(mod):
         setattr(this_module, setting, value)
 
 # TODO backward compatibility setting. Remove in Tethys 5.0
-USE_OLD_WORKSPACES_API = portal_config_settings.pop("USE_OLD_WORKSPACES_API", True)
+USE_OLD_WORKSPACES_API = portal_config_settings.pop("USE_OLD_WORKSPACES_API", False)
```
swainn, line 626: We can't make this change until 5.0 b/c it is a breaking change.

swainn (changes requested):
Please look into a solution for both the new workspace API and the old workspace API. we need to support both until 5.0

swainn (commented):
Does this fix work for both the old and new workspace APIs?

tethys_apps/base/paths.py
```diff
@@ -247,6 +246,9 @@ def _resolve_app_class(app_or_request, bypass_quota=False):
     ):
         app = app_or_request
     elif isinstance(app_or_request, HttpRequest):
+        # OLD VERSION
+        # app = get_active_app(app_or_request, get_class=False)
+        # NEW VERSION
```
swainn, line 251: Remove these comments

swainn (approved):

swainn:
@sdc50 I'd like for you to review this PR. It involves fixes for bugs in the Quotas API that were introduced, or at least perpetuated by the Paths API changes.

swainn:
@jakeymac can you rebase this branch with main?

swainn:
@jakeymac we'd like you to remove the bypass_quota parameter from the public workspaces API and use the private versions of the API in the user and admin views to bypass the quotas.

PR #1163, opened 2025-03-04
Jest testing

jakeymac:
This pull request adds front-end testing for the Tethys Platform.


tethys_gizmos/static/tethys_gizmos/js/slide_sheet.js
```diff
@@ -21,6 +21,9 @@ var SLIDE_SHEET = (function() {
     // Slide sheet
     var open_slide_sheet, close_slide_sheet;
 
+    // Private functions
+    var open, close;
```
jakeymac, line 25: Added these because in testing teardowns, leaving close as part of the global scope was causing some errors

PR #1174, opened 2025-06-02
Refactor the liveness-probe.sh to be able to check endpoints

swainn:
The current `liveness-probe.sh` script in the Docker only checks for process ids to verify that the processes are running. However, this does not verify if the endpoints are responding. I have a case where it would be helpful to verify the latter--the process is locked up, so still running (has a pid) but the endpoints timeout/don't respond.

This PR adds a `--check-endpoints` flag to the `liveness-probe.sh` script that when enabled checks the nginx and asgi endpoints with curl. The behavior is to pass if anything is returned regardless of http error code. It only fails if the endpoint times-out or doesn't return anything in 5s.

### Description
This merge request addresses...

### Changes Made to Code
 - Adds `--check-endpoints` flag to `liveness-probe.sh`
 - Adds `--debug` flag to `liveness-probe.sh`
 - If enabled, CHECK_ENDPOINTS mode will use `curl` to check the asgi endpoint (http://localhost:$TETHYS_PORT) and nginx endpoint (http://localhost:$NGINX_PORT). Any HTTP response will cause the test to pass regardless of HTTP code. Failure to return a response within 5 seconds will cause the test to fail.
 - If enabled, DEBUG mode sets `set -x` to print out all lines with arguments for debugging purposes.

- N/A

### Quality Checks

### DIFF
--- a/docker/liveness-probe.sh
+++ b/docker/liveness-probe.sh
@@ -1,30 +1,63 @@
 #!/bin/sh
 
+# liveness-probe.sh
+# Usage: ./liveness-probe.sh [--check-endpoints|-c] [--debug|-d]
+#
+# This script checks the health of supervisord, nginx, and tethys asgi server processes.
+# If the --check-endpoints or -c flag is provided, it will also check if the nginx and asgi endpoints respond.
+#
+# Exit code 0: All checks passed
+# Exit code 1: One or more checks failed
 
-check_process_is_running() { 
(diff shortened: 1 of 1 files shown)


swainn:
@gronka since you had a hand in making this file, I'd be interested in your thoughts on these changes.

gronka:
@swainn Looks like a solid addition. `curl` might need to be added to some requirements lists

PR #1175, opened 2025-06-05
Fix for Pip Install Overrides Django Pinning

swainn:
### Description
The PR #1169 introduced the dependencies into the pyproject.toml to support building a PyPI package in the future. However, these changes inadvertently override the pinned versions in the Docker image that is built. This PR fixes this issue. 

image

### Changes Made to Code
 - Run the pip install in the Dockerfile without installing dependencies--they are currently installed via the environment.yml.
 - Change ADD to COPY commands per docker best practices

- #1169 

### Additional Notes
-  In the future we will probably want to support building the docker image using pip or UV to install dependencies instead of conda, but this patch fixes it for the current approach of using Conda.

### Quality Checks


shawncrawley (approved):
Thanks for tackling this - good solution for now!

PR #1230, opened 2026-01-11
Fix active_venv extraction logic

shawncrawley:
Modify active_venv retrieval to only remove parenthesis if they are actually present.

### Description
This merge request addresses a small bug I found when using venv instead of conda where an assumption was made that the `VIRTUAL_ENV_PROMPT` environment variable would be surrounded in parenthesis. This assumption was made by me during my experience testing with using venv on a different laptop that did add parenthesis for some reason. Apparently that's not the standard and only behavior.

### Changes Made to Code
 - A check was added to see if the `VIRTUAL_ENV_PROMPT` environment variable starts and ends with a parenthesis rather than assume they are they and then stripping the first and last characters.

- 

### Additional Notes
-  

### Quality Checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
     "django-bootstrap5",
     "django-model-utils",
     "django-guardian",
+    "djangorestframework_simplejwt",
 ]
 
 [project.optional-dependencies]


shawncrawley:
I marked this as a draft PR because I still need to write a new test for the new logical path introduced.

PR #1265, opened 2026-04-13
Added Missing Dependency

jakeymac:
### Description
This merge request addresses a missing dependency needed for Tethys Component App development: "pyproj"

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -29,6 +29,7 @@ dependencies:
   - tornado>=6.5,<7
   - cryptography>=46.0.5
   - bcrypt # also required by channels, docker, daphne, condorpy
+  - pyproj
 
   # Gen CLI commands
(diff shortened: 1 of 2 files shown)


swainn:
@shawncrawley is pyproj a dependency of reactpy features in Tethys? or the specific application developed in the tutorial? I don't think ti should be a dependency of Tethys Platform.

swainn:
I think the better solution would be to make pyproj an optional dependency where it is referenced in the component utils and document it like we do with other optional dependencies. Pyproj is a tricky dependency that I don't want to dill with.

swainn:
We won't add `pyproj` as a dependency--it's a gnarly one. Instead, this dependency will be made optional and added to the documentation as an optional dependency by @shawncrawley in #1268

PR #1277, opened 2026-05-28
Secure Map Service

jakeymac:
### Description
This PR adds **Secure Map Services** to Tethys, enabling apps to register and consume map/image services that require authentication without exposing credentials to the browser.

The new feature supports:
- **API key** authentication with encrypted storage
- **OAuth2** authentication tied to a user’s linked social auth account
- optional **proxying** through Tethys Portal so credentials never reach the client
- retrieving secure services as a **map layer**, **endpoint**, or **server-side response**
- updating service parameters dynamically from app code

It also adds the supporting portal/admin UI, app-setting APIs, middleware for OAuth access requirements, CLI support for generating and managing a `SALT_KEY`, documentation, and tests.

### Changes Made to Code
- Added a new `SecureMapService` model in `tethys_services` for registering secure external map services
- Added `SecureMapServiceSetting` support in `tethys_apps` so apps can bind to secure services
- Added app-level APIs for:
  - declaring secure map service settings
  - retrieving a secure map service
  - updating service parameters
- Added a secure proxy endpoint for browser-safe access to authenticated services
- Added middleware to enforce app-specific OAuth account-link requirements
- Added admin forms/UI for creating and editing secure map services
- Extended map layout and gizmo support for token-aware map layers and GML layers
- Added CLI support for `SALT_KEY` generation and portal config bootstrapping
- Added documentation and a tutorial for using Secure Map Services
- Added tests for the new models, views, middleware, CLI commands, and map behavior

- None

### Additional Notes
- This PR introduces encryption support for stored API keys, so `SALT_KEY` must be configured in `portal_config.yml`.
- OAuth-based secure services may require users to link the corresponding provider account before accessing the app.

### Quality Checks
- [x] At least one new test has been written for new code
- [x] New code has 100% test coverage
- [x] Code has been formatted with Black
- [x] Code has been linted with flake8
- [x] The documentation has been updated appropriately


environment.yml
```diff
@@ -70,6 +70,7 @@ dependencies:
   - django-json-widget # enable json widget for app settings
   - djangorestframework # enable REST API framework
   - djangorestframework_simplejwt # JWT authentication for REST API
+  - django-fernet-encrypted-fields # enable encrypted fields for models
```
swainn, line 73: Check to see if this packages is on conda-forge. If it is not, we will need to create a feedstock for it and get it there.

tethys_apps/models.py
```diff
@@ -1104,6 +1116,192 @@ def create_persistent_store_database(self, refresh=False, force_first_time=False
...
+        """
+        Validate prior to saving changes.
+        """
+        if not self.secure_map_service and self.required:
+            raise ValidationError("Required.")
+
+    def _generate_request(self, param_overrides=None):
+        """
+        Generate a request to the secure map service, including any necessary authentication headers or parameters.
+        """
```
jakeymac, line 1164: @swainn I just noticed this function doesn't handle the OAuth2 token if the service setting is not set to use the proxy endpoint... Thoughts on how that should be handled?

jakeymac: If we want to worry about the OAuth token here at all, maybe we have a note/warning to just use the service as a layer or a response instead of getting the endpoint url without the OAuth token, or should this function perhaps return the OAuth token as well as the url

tethys_services/models.py
```diff
@@ -402,3 +404,112 @@ def get_engine(self, spatial=False):
...
+                "Authentication method must be 'oauth' to retrieve an OAuth token."
+            )
+        if not self.oauth_provider:
+            raise ValueError(
+                "OAuth provider must be specified to retrieve an OAuth token."
+            )
+
+        try:
+            auth = user.social_auth.get(provider=self.oauth_provider)
+        except ObjectDoesNotExist:
```
jakeymac, line 472: @swainn Right now if the Tethys Portal is not configured to require a user to be linked to an oauth in order to even access the app and the user's account is not linked to the necessary oauth service, this will crash the app, showing an error page with this error message. We should maybe consider handling this differently and/or adding instructions to add that requirement to the portal config.

docs/conf.py
```diff
@@ -93,7 +95,6 @@
     "sqlalchemy",
     "sqlalchemy.orm",
     "tethys_apps.harvester",
-    "tethys_apps.models",  # Mocked to prevent issues with loading apps during docs build.
```
swainn, line 96: Why was this removed?

tests/unit_tests/test_tethys_apps/test_views.py
```diff
@@ -291,3 +293,166 @@ def test_send_beta_feedback_email_send_mail_exception(
...
+        mock_service.authentication_method = "oauth"
+        mock_service.get_oauth_token.return_value = None
+        mock_get.return_value = mock_service
+
+        ret = secure_map_proxy(mock_request, mock_setting_id)
+
+        assert ret.status_code == 500
+        assert ret.content == b"Failed to retrieve OAuth token."
+
+    @mock.patch("tethys_apps.models.requests.request")
```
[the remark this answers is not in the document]
swainn: @jakeymac I agree with this one.

tethys_apps/base/app_base.py
```diff
@@ -1861,6 +1892,127 @@ def persistent_store_exists(cls, name):
...
+        param_overrides=None,
+        request_user=None,
+    ):
+        """
+        Retrieves secure map service assigned to named SecureMapServiceSetting for the app.
+
+        Args:
+            name(str): name of the SecureMapServiceSetting as defined in the app.py.
+            as_endpoint(bool): Returns endpoint url string if True, Defaults to False.
+            as_layer(bool): Returns GeoServerLayer object if True, Defaults to False.
```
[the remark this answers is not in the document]
swainn: Agree

tethys_services/models.py
```diff
@@ -402,3 +404,112 @@ def get_engine(self, spatial=False):
...
+        verbose_name_plural = "Secure Map Services"
+
+    def __str__(self):
+        return self.name
+
+    @classmethod
+    def get_authentication_method_options(cls):
+        """
+        Get the available authentication method options for the SecureMapService model.
+        This method is used for populating the choices in the admin form.
```
swainn: Are admin forms not able to handle choices on their own? It seems like Django would have figured this out.

tethys_services/models.py
swainn: Re: oauth is not setup on the portal - this is an optional requirement that should be flagged in the documentation. It should be called out in a tip box like we do with our optional dependencies.

Re: account not linked with oauth service - I would recommend redirecting them to their user profile settings page, which has settings for linking their account with enabled oauth services. Add a message to the redirect using the django messages API instructing them that they need to link their account with X provider. This will appear as a notification message on the user settings page.

tethys_services/models.py
```diff
@@ -402,3 +404,112 @@ def get_engine(self, spatial=False):
...
+        try:
+            auth = user.social_auth.get(provider=self.oauth_provider)
+        except ObjectDoesNotExist:
+            raise ValueError(f"User not linked to {self.oauth_provider}.")
+
+        access_token = auth.extra_data.get("access_token")
+        if not access_token:
+            raise ValueError("No access token found for user.")
+
+        return access_token
```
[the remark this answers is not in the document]
swainn: agree

tethys_apps/models.py
```diff
@@ -1104,6 +1116,189 @@ def create_persistent_store_database(self, refresh=False, force_first_time=False
...
+        resp = requests.get(service.endpoint, params=params, headers=headers)
+        if not resp.ok:
+            log.error(
+                f"SecureMapService with name {service.name} request failed: \n"
+                f"url: {service.endpoint}\n"
+                f"params: {params}\n"
+                f"headers: {headers}\n"
+                f"status_code: {resp.status_code}\n"
+                f"response_text: {resp.text}"
+            )
```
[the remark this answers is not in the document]
swainn: agree

tethys_apps/models.py
swainn: Handle this through documentation:
* If they want to use the as_endpoint without proxy, they need to get the token themselves
* Otherwise, recommend using the as_layer route.

tethys_apps/models.py
```diff
@@ -1104,6 +1116,189 @@ def create_persistent_store_database(self, refresh=False, force_first_time=False
...
+
+        headers = {}
+        if service.authentication_method == "oauth":
+            if not request_user:
+                raise ValueError(
+                    "Request user must be provided to fetch response for OAuth authenticated service."
+                )
+            headers["Authorization"] = f"Bearer {service.get_oauth_token(request_user)}"
+
+        resp = requests.get(service.endpoint, params=params, headers=headers)
```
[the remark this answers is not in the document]
swainn: agree

tethys_gizmos/templatetags/tethys_gizmos.py
```diff
@@ -147,9 +149,12 @@ def return_item(container, i):
         return None
 
 
-def json_date_handler(obj):
+def json_data_handler(obj):
```
swainn: You should probably make this a private function since you are renaming it:


```suggestion
def _json_data_handler(obj):
```

tethys_services/migrations/0005_alter_securemapservice_legend_title_and_more.py
swainn, line 1: Please merge the three migrations for securemapservice into a single migration. There are django migration commands to help do this.

tethys_services/static/tethys_services/js/secure_map_service_admin.js
```diff
@@ -0,0 +1,25 @@
...
+            apiKeyRow.style.display = ""
+            oauthProviderRow.style.display = "none"
+        } else if (method === "oauth") {
+            console.log("OAuth authentication selected")
+            oauthProviderRow.style.display = ""
+            apiKeyRow.style.display = "none"
+        } else {
+            console.log("No authentication selected")
+            apiKeyRow.style.display = "none"
+            oauthProviderRow.style.display = "none"
```
swainn: Remove the console.log() statements used for debugging.

tethys_services/backends/grid.py
```diff
@@ -0,0 +1,33 @@
...
+    ACCESS_TOKEN_METHOD = "POST"
+
+    REDIRECT_STATE = False
+
+    DEFAULT_SCOPE = ["api"]
+
+    SCOPE_SEPARATOR = ","
+
+    def user_data(self, access_token, *args, **kwargs):
+        return {"access_token": access_token}
```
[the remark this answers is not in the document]
jakeymac: GRiD does not provide any user data with Oauth2 or through its API. I have updated `user_data` and `get_user_details` to return empty dictionaries and added comments that clarify this.

tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
```diff
@@ -1017,12 +1073,101 @@ ol_layers_init = function()
...
+                let sep = baseUrl.indexOf('?') === -1 ? '?' : '&';
+                let url = baseUrl + sep + 'bbox=' + extent.join(',') + ',EPSG:3857';
+
+                fetch(url, {credentials: 'same-origin', headers: get_token_headers(token)})
+                  .then(r => r.text())
+                  .then(text => {
+                    let features = gmlFormat.readFeatures(text, {
+                      dataProjection: 'EPSG:4326',
+                      featureProjection: DEFAULT_PROJECTION,
+                    });
```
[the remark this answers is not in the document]
swainn: Agreed. See if there is a readProjection() method on the GML format to get the declared projection like we do with GeoJSON.

tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
```diff
@@ -1017,12 +1073,101 @@ ol_layers_init = function()
...
+            current_layer_layer_options['source'] = gmlSource;
+            layer = new ol.layer.Vector(current_layer_layer_options);
+          }
+
+          else if (current_layer.options.hasOwnProperty('gml')) {
+              let gmlSource = new ol.source.Vector({
+                  features: gmlFormat.readFeatures(current_layer.options.gml, {
+                      dataProjection: 'EPSG:4326',
+                      featureProjection: DEFAULT_PROJECTION,
+                  }),
```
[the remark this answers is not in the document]
swainn: Same comment: look for a way to read the projection from the GML.

tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
```diff
@@ -2313,6 +2458,72 @@ is_defined = function(variable)
...
+        img.src = object_url;
+      })
+      .catch(err => {
+        console.error('Image load failed: ', err);
+        if (is_defined(ol.ImageState) && image.setState) {
+          image.setState(ol.ImageState.ERROR);
+        }
+      });
+  };
+};
```
swainn: These functions are 66.6% the same. Can we combine them or pull out the common code into a separate function to make it more DRY?

tethys_gizmos/static/tethys_gizmos/js/tethys_map_view.js
```diff
@@ -911,9 +920,15 @@ ol_layers_init = function()
 
       // Tile layer case
       if (in_array(current_layer.source, TILE_SOURCES)) {
-        var resolutions, source_options, tile_grid;
+        var resolutions, source_options, tile_grid, tile_token;
+
+        tile_token = current_layer.options ? current_layer.options.token : null;
```
swainn, line 925: Copilot raised a couple of concerns about how we are handling tokens here. I'm most concerned with #1. But take a look at the others.

1. **No token rotation or refresh logic**: Once a token is embedded in the client-side options, there's no mechanism to refresh or rotate it. If a token is leaked or compromised, it remains valid until it expires on the server.

2. **Token exposed via console logs**: Tokens in Authorization headers are visible in console logs (potentially in the error messages that include URLs or status codes).

3. **Tokens in `current_layer.options` persist in the frontend state**: The token is extracted from options but never explicitly cleared from memory. Long-lived frontend state that contains tokens increases the window for exposure.

4. **GML/KML/Vector loaders use custom fetch with tokens**: Custom loader functions (e.g., lines 1019–1115 for GML) implement their own retry/error handling logic without rate-limiting, which could be exploited if a token is compromised.

I had one idea that would address both 1 and 3: Instead of passing the token as a layer option, just pass a flag that indicates if a token is required/available. Then setup an endpoint that the frontend can call to get a token for a given layer (using POST for most security). You can also handle the refresh logic with that endpoint.

tethys_cli/settings_commands.py
swainn, line 1: Add support to the `tethys services`, `tethys app_settings`, and `tethys link` commands to make it possible to create and assign secure map services programmatically through the CLI.

swainn (changes requested):

docs/conf.py
jakeymac: Mocking tethys_apps.models was causing classes that were documented using `autoclass` to render like this:

image

tethys_apps/models.py
```diff
@@ -1104,6 +1116,189 @@ def create_persistent_store_database(self, refresh=False, force_first_time=False
...
+        if not self.secure_map_service:
+            raise TethysAppSettingNotAssigned(
+                f"Cannot fetch response for SecureMapServiceSetting "
+                f'"{self.name}" for app "{self.tethys_app.package}": '
+                f"no SecureMapService assigned."
+            )
+        service = self.secure_map_service
+        params = dict(service.get_resolved_params() or {})
+        if param_overrides:
+            params.update(param_overrides)
```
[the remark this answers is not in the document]
jakeymac: Moved this logic to ``get_resolved_params``, a method on the service class itself. See tethys_services/models.py lines 501-506

environment.yml
jakeymac: Yes, the package is on conda-forge

tethys_services/models.py
jakeymac: Good catch. I think I was having issues getting this admin form field to populate before and this was a fix. After testing without this method, this doesn't seem to be needed anymore.

tethys_portal/middleware.py
```diff
@@ -155,3 +157,40 @@ def __call__(self, request):
         response = self.get_response(request)
 
         return response
+
+
+class TethysOauthRequiredMiddleware:
```
jakeymac: @swainn Perhaps this could be configured in the app.py file instead of inside portal_config.yml. Thoughts?

PR #1294, opened 2026-08-04
Let job status be reported to the portal instead of polled for

romer8:
## Why

The jobs table poll views refresh job status on every request. For condor jobs that opens an SSH connection to the scheduler inside the request/response cycle, and rendering a workflow's DAG costs one remote query *per node*, measured at ~115 ms per expanded node.

Median latency of the DAG-diagram poll on 26-node DAGs against a real pool, caller pinned to 0.5 CPU, L concurrent viewers each watching their own workflow, zero failures. Two independent runs of this change, to show the spread:

| Viewers | released | + faster condorpy | + this change | (2nd run) |
|---|---|---|---|---|
| 2 | 7000 ms | 2400 ms | **300 ms** | 330 ms |
| 4 | 12000 ms | 6300 ms | **440 ms** | 530 ms |
| 6 | 18000 ms | 5000 ms | **890 ms** | 930 ms |
| 8 | 26000 ms | 6100 ms | **1400 ms** | 1400 ms |

Throughput at 8 viewers: 0.50 → 4.63 and 4.58 req/s. In both runs all 10 workflows were served from the database for the whole run, rather than falling back to reading live.

The middle column is with a condorpy that caches decrypted keys, reuses SSH clients, and answers `Workflow.statuses` in one query. That leaves the DAG diagram reading each node individually. At 2 viewers it costs the same as an isolated request, so it is a per-poll floor rather than contention: more CPU or replicas cannot help, and it grows with the DAG.

**This branch does not depend on that condorpy.** An earlier revision had an `update_node_statuses()` that pulled statuses through `node_statuses_by_cluster_id`, which no released condorpy has; it has been removed. Everything here uses `CONDOR_JOB_STATUSES` and `update_node_ids`, both in 0.6.0. The measured numbers come from the push path and are unaffected. Key caching and client reuse are still worth having on their own merits, since they speed up the live fallback.

`update_row` reached the scheduler by a second path through `job.statuses`, so gating only the status refresh would not have removed remote calls from the poll.

## What this changes

**Serve from the database.** `CondorWorkflowNode.cached_node_status` holds each node's last known status. `apply_node_statuses()` persists a reported mapping and returns what it matched. `node_statuses_updated` records when a report last matched at least one node, and `node_statuses_are_current` reports whether that was within `node_statuses_max_age`. The views render from the database when it is, and read live when it is not. `cached_node_statuses` reports counts in the same shape as `CondorBase.statuses`, sourced from the node rows so no remote call is made.

Both DAG sources go through one builder. `get_condor_job_nodes` previously had two branches independently assembling the same dict, and they disagreed: the database branch keyed on the raw node name while the live branch keyed on the CondorPy job name, which differ for any node whose name contains a space. Each source now yields `(name, status, cluster_id, parents)` and one comprehension turns that into the response, so they cannot drift and both key on the CondorPy name. Parents resolve through a `{pk: name}` map built from the nodes already being iterated, rather than duplicating CondorPy's underscore rule here.

**Accept reports rather than only asking.**

```
POST /report-job-status/<job_id>/?token=<token>&status=Complete
                                &node_statuses={"prep":"Completed",...}
```

`update_job_status()` already exists, but it asks the job's *source* — for condor the portal SSHes out to rediscover what it was just told. This records what it is given, so the party that already knows can do the telling. For condor that is best placed alongside the scheduler, where the queue is a local query and where evictions, holds and memory kills are visible; a job cannot report those about itself because it is no longer running.

Reports are authorised by `TethysJob.status_report_token`, signed with the portal's `SECRET_KEY`. Nothing to configure, no secret to distribute, and a token authorises one job and no other. Tokens expire after `STATUS_REPORT_MAX_AGE` (30 days) — generous, because a token has to outlive a running job; the point is to bound how long one stays usable after being read off the queue.

A refused report answers with `success: false`, an `error`, and one of `403` (token missing, malformed, expired, or for another job), `400` (unrecognised status), `404` (no such job), or `409` (see below).

**Reported values are treated as remote input.** A status condor does not define is logged and ignored rather than persisted, and the read side defaults instead of indexing `STATUS_MAP` — a reporter sending `{"prep": "Banana"}` previously had that stored, and the DAG view then raised `KeyError` on every poll until the statuses went stale, blanking the diagram for everyone watching. A report matching no node no longer stamps the freshness timestamp, which used to leave the views serving a DAG of unexpanded nodes at 0% as though it were fresh. Matched nodes are written with one `bulk_update` inside a transaction rather than a save each, since a reporter re-sends them on an interval.

**A report may not revive a finished job.** Doing so is not just a wrong badge: the next poll re-reaches the terminal status with a non-terminal one recorded, so `update_status` treats it as a fresh completion and processes the results a second time, re-syncing the remote output. Those reports are refused with `409`. A repeated terminal status is still accepted, deliberately — a reporter retries after a timeout, and `Various-Complete` is itself terminal and is what a workflow reaches on its way to `Complete`. Recognising a finished job goes through a new `is_terminal`, because a custom terminal status registered with `add_custom_terminal_status` is stored as `OTH` and a check against `TERMINAL_STATUS_CODES` alone misses it.

**A failed result sync can be retried.** A terminal status is recorded even when the work that follows it fails, so a transient problem does not leave the portal believing a finished job is still running; the response reports `post_processing: false` rather than `success: false`. That used to mean such results could never be fetched: the job was terminal, `update_needed` was `False` forever, and no retry existed. This needed no new state — `process_results` stamps `completion_time` only once `_process_results` has returned, so a job in a results status without one is exactly one that did not get through it (`post_processing_incomplete`). Being told the status again now retries that work, rate-limited like a refresh so repeated reports for a job whose sync keeps failing cannot each start their own, and never on a poll, which must not pay for an SCP inside a request.

**Tell the reporter which job it is.** `CondorWorkflow.status_report_options` attaches the job id and its report token to the DAGMan job as ClassAds:

```
-append '+TethysJobId = "42"'
-append '+TethysJobToken = "<signed token>"'
```

Without this the endpoint has no practical caller: whatever watches the queue can see the DAG but has no way to say which `TethysJob` it belongs to, or to prove it may report on it. Doing it here rather than in each application keeps the signing secret inside the portal and means every deployment gets it without its own change. The ads are inert where nothing is reporting. The ad names are constants (`JOB_ID_AD`, `JOB_TOKEN_AD`) so a rename is visible in one place, and they are shell-quoted only for a remote submit, which condorpy hands to a shell — quoting the local path, which goes through argv, would make condor read the quotes as part of the value.

## Decisions worth flagging

**The staleness budget is its own value.** `node_statuses_max_age` (60 s, overridable like `update_status_interval`) is deliberately not the poll interval. That interval is the minimum time between polls of a *single* job; this is how stale a *rendered* status may be. Anything refreshing many jobs takes longer per pass than the interval for one job, so tying them together leaves statuses never current exactly when load makes it matter — measured: a pass over 18 jobs took 16.8 s against the 10 s interval, and every workflow fell back to live reads.

**Freshness rather than a setting.** A deployment where nothing reports never has current statuses, so it reads live and behaves as before, with nothing to configure. And if a reporter stops, the views resume reading live rather than serving values that never advance — a setting would leave a frozen jobs table that looks like working software.

**A reporter must re-send node statuses on an interval, even when nothing changed.** Sending only on change lets them expire on exactly the steady-state workflows the cache is meant to help. This is now written into the documentation rather than being something the next reporter author rediscovers.

## Also

`TethysJob._last_status_update` becomes a persisted field. It was set only in memory, so a fresh instance per request always fell back to `execute_time` and `is_time_to_update()` always returned `True` — the interval never throttled anything across requests. The property no longer assigns while being read. This means `update_status_interval` now genuinely limits refreshes for **all** job types, which is a behaviour change for any app that relied on a status refresh per request; `whats_new` says so.

Both node-status paths key on the condorpy job name, not the raw node name: `CondorPyJob.condorpy_job` builds it as `self.name.replace(" ", "_")`, so keying on `node.name` silently fails to match any workflow with spaces in its node names — no exception, just a DAG reporting every node unexpanded.

`get_job_statuses` keeps its existing bucketing, including the pre-existing quirk that only `Completed` and `Running` are ever populated, so this PR only moves where the numbers come from. Its denominator comes from whichever source the counts did, because reading `num_jobs` on the database path would build the condorpy workflow and reopen the connection this path exists to avoid.

## Documentation

The endpoint, its parameters, response envelope and every refusal code are documented next to the existing `update-job-status` callback in `jobs.rst`, along with how to obtain a token, its lifetime, and the fact that anything holding one can set that job's status. The ClassAd protocol and the periodic-reporting requirement are documented with the Condor workflow job type. `whats_new` covers both the endpoint and the `_last_status_update` behaviour change, and notes that migrations must be applied before serving with the new code since one of the new columns is on the base job table.

## Testing

62 new tests: 23 in `test_CondorWorkflow.py`, 21 in `test_update_status.py`, 14 in `test_TethysJob.py`, 4 in `test_jobs_table.py`. They cover the freshness property in five states, bulk persistence, the spaces-in-name case, a report matching nothing, an unrecognised status, both DAG sources producing the same shape, and the authorisation boundary — missing token, forged token, a valid token for a *different* job, invalid status, unknown job, the 409 on reviving a finished job, a terminal status surviving a failure in the work that follows it, and that failure being retried by a later report but not by a poll. The submit ads are covered by asserting the token verifies back to the job that produced it, no longer verifies once tampered with, and is quoted only for a remote scheduler.

Some of these go through the URL rather than calling the view, which the earlier ones did not, and which mattered: `csrf_exempt` wraps an async view in a *sync* function on Django 4.2, so the endpoint returned an unawaited coroutine and every report failed with a 500 while the view's own tests passed. It is a sync view now, like `update_dask_job_status` beside it.

Migration `0003` adds the three fields; `makemigrations --check` clean, as are flake8 and black. Full suite locally: 2356 passing, with only the pre-existing `test_tethys_cli` pair failing, confirmed by reproducing them at the previous head. Verified end to end against a real pool with a reporter running beside the scheduler: `Submitted → Various (+ node statuses) → Complete`, and all 10 workflows served from the database for the duration of a load run.

## Known issue, deliberately not fixed here

`update_status` ends with a whole-row `save()`, and this PR adds a second writer, so a poll that loaded a job before a report landed can write its stale values back — including `node_statuses_updated`. The obvious fix, restricting that save with `update_fields`, is worse than the bug: `CondorBase.update_database_fields()` does `self.remote_id = self.remote_id or self._condor_object._remote_id`, and `remote_id` reaches the database only because some full save happens to carry it. Restrict it and the next request generates a different id, so commands run in a remote directory that does not exist. The payoff is small either way — the field regresses to a slightly older timestamp, the views read live briefly, and the next report restores it. Doing it properly means each class declaring which fields its `pre_save` hook manages, which is a change to shared save behaviour for every job type.

## Still true after this

No reporter ships in this repo. The protocol is documented, but the implementation lives outside it, so a deployment gets the fast path only by writing or installing one. And a token in a queue-readable ClassAd means status reporting is only as trustworthy as the pool, since a holder can mark a job complete and trigger its results processing. Bounded and documented now, not removed.


romer8:
Pushed three commits from a review pass over this branch. The first one changes what this PR depends on, so it is worth reading before the rest.

## This no longer waits on condorpy

`update_node_statuses` is gone. It was the only caller of `node_statuses_by_cluster_id`, which no released condorpy has, so on any deployment installing condorpy from conda-forge it could only raise inside its own `try`, log a warning, and return nothing. It also had no caller in this repo: it was written for an app-tier updater that was abandoned in favour of a reporter running beside the scheduler, which pushes through the endpoint into `apply_node_statuses` instead.

Nothing else here touches unreleased condorpy API — `CONDOR_JOB_STATUSES` and `update_node_ids` are both in 0.6.0 — so dropping it decouples this branch from condorpy#38 entirely. The measured numbers in the description come from the push path and are unaffected. condorpy#38 is still worth having on its own merits, since key caching and client reuse speed up the live fallback, but this can land without it.

## Reported node statuses are treated as remote input

They arrive from outside the portal, so the endpoint no longer takes them at face value.

A status condor does not define is ignored rather than persisted, and the read side defaults instead of indexing `STATUS_MAP` directly. Previously a reporter sending `{"prep": "Banana"}` had that value stored, and the DAG view then raised `KeyError` on every poll until the statuses went stale — caught by the existing handler, so not a 500, but the diagram went blank for everyone watching.

A report matching no node no longer marks the statuses current. The timestamp was stamped after the loop regardless, so a report whose node names matched nothing left the views serving a DAG of unexpanded nodes, at 0%, as though it were fresh.

Matched nodes are written with one `bulk_update` inside a transaction rather than a save each, since a reporter re-sends them on an interval.

## Both DAG sources go through one builder

`get_condor_job_nodes` had two branches independently building the same dict. They also disagreed: the database branch keyed on the raw node name while the live branch keyed on the CondorPy job name, which differ for any node whose name contains a space. Harmless today because the graph is rebuilt per poll and the labels come out the same, but it was a trap for anything that ever compares node identity across polls.

Each source now yields `(name, status, cluster_id, parents)` and one comprehension turns it into the response, so they cannot drift and both key on the CondorPy name. Parents resolve through a `{pk: name}` map built from the nodes already being iterated, rather than duplicating CondorPy's underscore rule here.

Also `cached_statuses` is now `cached_node_statuses` — it was one letter away from the unrelated job-level `cached_status` — and the percentages come from the counts already fetched instead of a second `COUNT(*)`.

## Reported status transitions are checked, and a failed sync can be retried

Two problems either side of the same line.

A report could move a job out of a terminal status. That is not just a wrong badge: the next poll re-reaches the terminal status with a non-terminal one recorded, so `update_status` treats it as a fresh completion and processes the results a second time, re-syncing the remote output. Those reports are refused with 409.

A repeated terminal status is still accepted, deliberately. A reporter retries after a timeout, and `Various-Complete` is itself terminal and is what a workflow reaches on its way to `Complete`, so refusing anything but an identical repeat would 409 the normal end of a workflow. Recognising a finished job goes through a new `is_terminal`, because a custom terminal status registered with `add_custom_terminal_status` is stored as `OTH` and a check against `TERMINAL_STATUS_CODES` alone misses it.

The opposite failure was that a terminal status is recorded before the work that follows it — on purpose, so a failed sync does not leave a finished job looking like it runs — which meant results that failed to sync could never be fetched. The job was terminal, `update_needed` was `False` forever, and no retry existed. This needed no new state: `process_results` stamps `completion_time` only once `_process_results` has returned, so a job in a results status without one is exactly one that did not get through it. Being told the status again now retries that work, rate-limited like a refresh so repeated reports for a job whose sync keeps failing cannot each start their own, and never on a poll, which must not pay for an SCP inside a request.

## Report tokens expire

`signing.loads` was called without `max_age`, so a token was valid forever and only rotating `SECRET_KEY` revoked one. They now expire after 30 days — generous, because a token has to outlive a running job; the point is to bound how long one stays usable after being read off the queue, where any user who can query the scheduler can see it. That exposure is inherent to carrying the token in a job ad and is now stated in the docs, along with the fact that CondorPy logs the full submit command at `INFO`, so its logger should not be turned up in production.

## Documentation

The endpoint, its parameters, response envelope and every refusal code are documented next to the existing `update-job-status` callback, and the ClassAd protocol is documented with the Condor workflow job type, with the ad names lifted into constants so a rename is visible in one place.

The part worth flagging: a reporter has to re-send node statuses **on an interval, even when nothing has changed**. Sending only on change lets them expire on exactly the steady-state workflows the cache is meant to help — which is how it behaved before a heartbeat was added on the reporter side. That is now written down rather than being something the next reporter author rediscovers.

`whats_new` also notes that persisting `_last_status_update` makes `update_status_interval` throttle every job type rather than only condor, which is a behaviour change for any app that relied on a refresh per request.

## One finding deliberately not fixed

`update_status` ends with a whole-row `save()`, and this PR adds a second writer, so a poll that loaded a job before a report landed can write its stale values back — including `node_statuses_updated`.

The obvious fix, restricting that save with `update_fields`, is worse than the bug. `CondorWorkflow` and `CondorJob` have `pre_save` receivers calling `update_database_fields()`, and `CondorBase.update_database_fields()` does `self.remote_id = self.remote_id or self._condor_object._remote_id`. `remote_id` is a fresh `uuid4` each time a remote client is built and reaches the database only because some full save happens to carry it, that one included. Restrict it and the next request generates a different id, so commands run in a remote directory that does not exist, output syncs from the wrong place, and `_log_files` looks where the logs are not.

The payoff is small either way: the window is between loading a job and saving it inside one request, and the field regresses to a slightly older timestamp, so the views read live briefly and the next report restores it. Doing it properly means each class declaring which fields its `pre_save` hook manages so they can be merged into `update_fields` — a change to shared save behaviour for every job type, which does not belong in this PR. Happy to be argued out of that.

## Still true after this

No reporter ships in this repo. The protocol is documented, but the implementation lives outside it, so a deployment gets the fast path only by writing or installing one. Worth deciding whether a reference implementation belongs here.

And a token in a queue-readable ClassAd means status reporting is only as trustworthy as the pool, since a holder can mark a job complete and trigger its results processing. Bounded and documented now, not removed.

## Verification

Full suite locally: 2356 passing, and the only two failures are the pre-existing
`test_tethys_cli` pair (`test_app_settings_gen_salt_strings_command_all_apps`,
`test_list_command_urls`) — confirmed pre-existing by stashing these changes and reproducing them
at the previous head. black, flake8 and `makemigrations --check` clean.

On CI, 36 checks pass, including every test job across Django 4.2 and 5.2 on Python 3.10–3.13 for
both Ubuntu and macOS, and coveralls.

Two checks are red and neither is from this push. The Read the Docs build was already failing on
the previous head (`626abff4`) before any documentation was touched here. For what it is worth,
the three `.rst` files this adds check out: each new `:ref:` target resolves to exactly one label,
there are no duplicate label definitions anywhere under `docs/`, heading underlines are all
correct, and the one simple table parses cleanly under docutils. Merge Gatekeeper ran while other
jobs were still queued and lists them as pending, so it is a timing artifact rather than a result.

swainn (approved):

swainn (approved):

PR #1305, opened 2026-09-22
Enable server-side revocation of JWT refresh tokens

Sonia-96:
### Description

Until now there was no way to invalidate a refresh token. Logout could not revoke it, so a refresh token kept minting new access tokens until its natural expiry. With the blacklist app and this endpoint, a client can surrender its refresh token on logout and it is rejected afterward.

### Changes Made to Code

Adds a way to revoke JWT refresh tokens server-side:
- New `/api/token/blacklist/` endpoint (simplejwt `TokenBlacklistView`).
- Installs the `rest_framework_simplejwt.token_blacklist` app.
- Defaults `SIMPLE_JWT` to `ROTATE_REFRESH_TOKENS` and `BLACKLIST_AFTER_ROTATION`, both overridable via `portal_config.yml`.

### Quality Checks
 - [x] New code has 100% test coverage
 - [x] Code has been formatted with Black
 - [x] Code has been linted with flake8
 - [x] The documentation has been updated appropriately


tethys_portal/urls.py
```diff
@@ -172,6 +171,9 @@
     ),
     re_path(r"^token/refresh/$", TokenRefreshView.as_view(), name="token_refresh"),
     re_path(r"^token/verify/$", TokenVerifyView.as_view(), name="token_verify"),
+    re_path(
+        r"^token/blacklist/$", TokenBlacklistView.as_view(), name="token_blacklist"
```
swainn, line 175: What is this view? Is it required add this view to use the blacklist app? Or is this just for convenience while developgin?

tethys_portal/settings.py
```diff
@@ -349,6 +350,17 @@
...
+# refresh token surrendered at /api/token/blacklist/ (e.g. on logout) is revoked
+# server-side and can no longer mint access tokens. Overridable via portal_config.yml.
+SIMPLE_JWT = portal_config_settings.pop(
+    "SIMPLE_JWT",
+    {
+        "ROTATE_REFRESH_TOKENS": True,
+        "BLACKLIST_AFTER_ROTATION": True,
+    },
+)
+
```
swainn: These settings could be set by a portal admin in their portal_config.yaml when they are installing their portal. I'm not sure it makes sense to have them as part of the default settings configuration.

swainn (changes requested):
Please create a guide for using JWT tokens in Tethys in the Manual Production Installation > Advanced Configuration > Optional section of the docs (https://docs.tethysplatform.org/en/latest/installation/production/manual/configuration.html#advanced-configuration). This should describe how to use the JWT feature in Tethys, what the default behavior is, and how to configure it with links to the Simple JWT documentation.

tethys_portal/urls.py
Sonia-96: `TokenBlacklistView` is the built-in `simplejwt` view that lets a client revoke its own refresh token (`POST /api/token/blacklist/` with the refresh token). It isn't strictly required for the blacklist app to work. The app already blacklists old refresh tokens automatically on each rotation because we set `ROTATE_REFRESH_TOKENS` and `BLACKLIST_AFTER_ROTATION`. What the URL adds is on-demand revocation: a client can invalidate a token immediately on logout or if it's compromised, instead of it staying valid until it expires. It's a standard production endpoint, not a dev-only convenience. If we don't want to expose it, we can drop the URL and rotation-based revocation still works.

